PR #20, opened 2016-04-12
Fix crash caused by recreating static nested classes using reflection

adi1133:
> Caused by: java.lang.RuntimeException: An exception occurred while creating a new instance of de.myapp.ui.user.UserInfoActivity.MyChangeHandler. An exception occurred while finding class for name de.myapp.ui.user.UserInfoActivity.MyChangeHandler. de.myapp.ui.user.UserInfoActivity.MyChangeHandler

The same error also occurs with Controllers
Relevant [SO](http://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav) post


EricKuck:
👍 Great catch, thanks!

Issue #17, opened 2016-04-11
Facebook integration


leonardo2204:
I'm implementing a Facebook login implementation on top of the demo app, but I can't find a way to bind the onActivitityResult from Facebook to my Controller. 
What I have:
```
/**
 * Created by leonardo on 4/11/16.
 */
public class FacebookController extends BaseController {

    @Bind(R.id.login_button)
    LoginButton loginButton;

    private CallbackManager callbackManager;

    @Override
    protected View inflateView(@NonNull LayoutInflater inflater, @NonNull ViewGroup container) {
        return inflater.inflate(R.layout.controller_facebook, container, false);
    }

    @Override
    protected void onViewBound(@NonNull View view) {
        super.onViewBound(view);

        callbackManager = CallbackManager.Factory.create();
        loginButton.registerCallback(callbackManager, new FacebookCallback<LoginResult>() {
            @Override
            public void onSuccess(LoginResult loginResult) {
                Toast.makeText(getActivity(), "Logged in !", Toast.LENGTH_SHORT).show();
            }

            @Override
            public void onCancel() {
                Toast.makeText(getActivity(), "Canceled by the user!", Toast.LENGTH_SHORT).show();
            }

            @Override
            public void onError(FacebookException error) {
                Toast.makeText(getActivity(), "Error:"+ error.getMessage(), Toast.LENGTH_SHORT).show();
            }
        });
        loginButton.setOnClickListener(null);
    }

    @OnClick(R.id.login_button)
    public void onClick(View v){
        LoginManager.getInstance().logInWithReadPermissions(getActivity(), null);
    }

    @Override
    public void onActivityResult(int requestCode, int resultCode, Intent data) {
        super.onActivityResult(requestCode, resultCode, data);
        callbackManager.onActivityResult(requestCode, resultCode, data);
    }

    @Override
    protected String getTitle() {
        return "Facebook SDK Login";
    }
}
```
I realized that in order to get the Activity's lifecycle callbacks I must call the Controller startActivityForResult for instance, but as I'm using Facebook SDK I don't have the control over this.
How should I proceed in that case ?

Thanks

EricKuck:
[...]

PR #23, opened 2016-04-13
Automatically deploy snapshot from tavis

sockeqwe:
As you might have already noticed you are releasing quite often minor changes and bugfixes like #17 #20 etc.

This is not a bad thing!

But I would like to discuss, if it would make more sense to deploy snapshots to give those users that are really in need of the merged changes like #17 fast, but also gives you the ability to be a little bit slower regarding versioning and releasing. What I mean is, instead of publishing a new version every 2-3 days, publish a snapshot after each commit / pull request, and after having applied two or three changes publish a new version.

So in this pull request, I want to purpose to use travis ci to automatically publish SNAPSHOTS.

**Please don't merge this pull request for now**. There are some things that we have to discuss as I haven't used bintray before and probably I have to adjust some files and you also have to fill in some things like bintray credentials.

What do you think about deploying SNAPSHOTS automatically?

### DIFF
--- a/.buildscript/deploy_snapshot.sh
+++ b/.buildscript/deploy_snapshot.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# Deploy a jar, source jar, and javadoc jar to Sonatype's snapshot repo.
+#
+# Adapted from https://coderwall.com/p/9b_lfq and
+# http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/
+
+SLUG="bluelinelabs/Conductor"
+JDK="oraclejdk8"
+BRANCH="develop"
+
+set -e
+
+if [ "$TRAVIS_REPO_SLUG" != "$SLUG" ]; then
+  echo "Skipping snapshot deployment: wrong repository. Expected '$SLUG' but was '$TRAVIS_REPO_SLUG'."
+elif [ "$TRAVIS_JDK_VERSION" != "$JDK" ]; then
+  echo "Skipping snapshot deployment: wrong JDK. Expected '$JDK' but was '$TRAVIS_JDK_VERSION'."
+elif [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  echo "Skipping snapshot deployment: was pull request."
+elif [ "$TRAVIS_BRANCH" != "$BRANCH" ]; then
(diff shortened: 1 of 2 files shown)


EricKuck:
Thank you! This is something we do internally with our internal GitLab and private maven repo setup, but that I haven't taken the time to do for open source stuff yet. This is perfect.

The downside of bintray is that they don't support snapshot releases, so I'll have to decide whether I should just go back to maven central or try to hook up jfrog oss. Happen to have any experience w/ jfrog at all?

sockeqwe:
Unfortunately, I have only worked with sonatype so far. 

> The problem is that Bintray does not support SNAPSHOTs (Bintray is for releases only). What you need is an Artifactory+Bintray combo, that deploys snapshots to Artifactory, and once in a while (when you deside), releases a 'production-ready' version to Bintray.
> 
> Depending on the nature of your project you might be qualified to a free account on oss.jfrog.org. The requirement is that your deliverable is an open-source library/product, which is included in JCenter.
> 
> Pushing from Artifactory to Bintray is a really simple process, you can perform it just by clicking a button in Artifactory UI or by performing a REST call. Just remember - it must be a release, not a SNAPSHOT.

So I guess you have to go the "work around" with jfrog oss if you want to use bintray.

I will do some google research to dive deeper into this topic.

adi1133:
Doesn't Jitpack fit this role ?

sockeqwe:
jitpack is great for small hobby projects but I wouldnt rely my business on it: what if jitpack will be shut down, general availability of jitpack (not mirrored), etc. 

I know, we are just talking about SNAPSHOTS , but ...

EricKuck:
Quick update: finally got approved to push to JFrog OSS, so I'll be playing with that for a bit. I can't find much documentation on how it works (and if it can automatically push to sonatype the way bintray can), so if it doesn't work out, I'll just use sonatype.

EricKuck:
I ended up ditching bintray, as the jfrog oss stuff ended up being pretty confusing. I switched over to sonatype and got snapshot releases building.

Unfortunately I missed a command when merging your PR and committed it as myself instead of actually doing a legit merge. The code's there, but it's credited to me instead of you. Sorry about that! If there's a way to fix it that I'm not aware of, please let me know.

sockeqwe:
No problem. I'm glad snapshot deployment works! Thanks!

PR #25, opened 2016-04-15
added methods to lifecyclelistener for saving/restoring viewstate

sockeqwe:
It would be nice to have lifecycle listeners for `onSaveViewState()` and `onRestoreViewState()`. Would make life easier for other third party plugins like Mosby.

Wanted to at tests for that, but haven't found tests for `onSaveInstanceState()` and `onRestoreInstanceState()` where I could add my tests for `onSaveViewState()` and `onRestoreViewState()`. Have I missed something?

Have I picked the right bundle for the view state?

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -1084,7 +1084,7 @@ public abstract class Controller {
         public void onRestoreInstanceState(@NonNull Controller controller, @NonNull Bundle savedInstanceState) { }
 
         public void onSaveViewState(@NonNull Controller controller, @NonNull Bundle outState) { }
-        public void onRestoreViewState(@NonNull Controller controller, @NonNull Bundle savedInstanceState) { }
+        public void onRestoreViewState(@NonNull Controller controller, @NonNull Bundle savedViewState) { }
     }
 
 }
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -907,6 +907,10 @@ public abstract class Controller {
                 child.controller.saveViewState(child.controller.mView);
             }
         }
+
+        for (LifecycleListener lifecycleListener : mLifecycleListeners) {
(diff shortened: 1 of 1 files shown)


adi1133:
Shouldn't all lifecycle listener methods have a pre and post callback?

sockeqwe:
`onSaveInstanceState()` doesnt have pre and post callbacks either. I guess @EricKuck  had good reasons to do so. In contrast to "real" lifecycle callbacks like `onAttach()` I don't think that you have to know "pre" or "post" for `onSaveInstanceState()` or `onSaveViewState()`

EricKuck:
The pre- and post- callbacks are for methods that can change a fully configured controller's state. For example, an outside observer may be interested i knowing that a controller is about to be attached so it can start some external animation (maybe the toolbar will appear/disappear). Another observer may only care that the controller has fully attached (maybe to make a non-animated toolbar title change).

Saving and restoring states are different, and therefore do not have pre- and post- callbacks. Saving state should never alter the controller, so pre- and post- would be superfluous. Restoring the state would alter the controller, but it's not even really a valid controller until it's had its state restored anyway. If people think pre- and post- would make sense here, I wouldn't be totally opposed to adding them. I just don't see a need.

conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
```diff
@@ -1074,6 +1082,9 @@ public void postDestroy(@NonNull Controller controller) { }
 
         public void onSaveInstanceState(@NonNull Controller controller, @NonNull Bundle outState) { }
         public void onRestoreInstanceState(@NonNull Controller controller, @NonNull Bundle savedInstanceState) { }
+
+        public void onSaveViewState(@NonNull Controller controller, @NonNull Bundle outState) { }
+        public void onRestoreViewState(@NonNull Controller controller, @NonNull Bundle savedInstanceState) { }
```
EricKuck: `savedInstanceState` should probably be `savedViewState` here.

EricKuck:
As for this PR, it looks good 👍. I made one SUPER nitpicky request and will merge over the weekend. I'll be building out more tests over the weekend as well, so I'll add some for this too. Thanks @sockeqwe!

sockeqwe:
:+1: done

PR #30, opened 2016-04-22
Make getControllerWithInstanceId search recursively

### LINKED ISSUE #29: getTargetController() returns null if target is nested too many times
getTargetController() returns null if target is nested too many times, this happens after process death.
I made a project to reproduce the issue.
https://github.com/adi1133/Conductor-Issue

The issue is caused by the fact that `getChildControllerWithInstanceId(instanceId)` is not recursive.

adi1133:
This PR adds a new public method to Controller `getControllerWithInstanceId(instanceId)` for searching for a controller recursively. This returns the controller itself, or any of its descendants (children).

This will fix #29 

This is currently an addition to the public api.
Should I alter this PR to do any of the following ?

- remove `getChildControllerWithInstanceId(instanceId)` 
- deprecate `getChildControllerWithInstanceId(instanceId)` 
- make `getChildControllerWithInstanceId(instanceId)`  search recursively and keep the existing api
- make the new method internal

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -271,17 +271,17 @@ public abstract class Controller {
     }
 
     /**
-     * Returns the a nested Controller with the given instance id, if available.
-     *
+     * Returns the Controller with the given instance id, if available.
+     * May return the controller itself or a matching descendant
      * @param instanceId The instance ID being searched for
      * @return The matching Controller, if one exists
      */
-    public final Controller getControllerWithInstanceId(String instanceId) {
-        if (getInstanceId().equals(instanceId))
+    final Controller findController(String instanceId) {
+        if (mInstanceId.equals(instanceId))
             return this;
 
         for (ControllerTransaction transaction : mChildControllers) {
-            Controller controllerWithId = transaction.controller.getControllerWithInstanceId(instanceId);
+            Controller controllerWithId = transaction.controller.findController(instanceId);
             if (controllerWithId != null)
                 return controllerWithId;
         }
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -270,6 +270,24 @@ public abstract class Controller {
     }
 
     /**
+     * Returns the a nested Controller with the given instance id, if available.
+     *
+     * @param instanceId The instance ID being searched for
+     * @return The matching Controller, if one exists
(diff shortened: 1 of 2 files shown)


EricKuck:
Thanks for this! I can't quite decide what the best approach to this problem would be. The two best options I see are:

- Make the existing `getChildControllerWithInstanceId(instanceId)` search recursively
  - __Pros__: 
    - Keeps the same API.
    - Controllers have unique IDs, so you'd never accidentally get the wrong controller.
  - __Cons__:
    - It would be weird for this to work recursively while `getChildController(String tag)` and `getChildControllers()` only return immediate children. The tag getter couldn't be made recursive, due to tags not being unique, and `getChildControllers()` shouldn't be recursive for obvious reasons.

- Make the new method internal
  - __Pros__: 
    - Stuff just keeps on working the way it was already working.
  - __Cons__:
    - Still confusing to have a new method that does _almost_ the same thing, just recursively, even if it is internal.

Anyone have thoughts on this? I'm leaning toward option 1 right now.

adi1133:
If we make the existing getChildControllerWithInstanceId(instanceId) search recursively, the method name might not be intuitive of its behaviour.

I sugest deprecating the old method and creating a new one named findControllerWithInstanceId or findController(String instanceId) or something similar.

Jquery naming example:
https://api.jquery.com/children/

EricKuck:
Is there a use for making this method public? I guess I can't think of anything it _should_ be used for from an app developer's perspective. I agree that we shouldn't change the existing method, but maybe it should be package-private.

adi1133:
In my opinion there are the following use cases:
- For referencing other arbitrary controller
  - For a single other controller: `setTarget()`, `getTarget()`
  - For multiple other controllers: `getRouter().getControllerWithInstanceId(instanceId)` and storing the instanceId string in a bundle.
- For referencing child controllers: `getChildController(tag)`

I am not sure if an app developer needs `getChildControllerWithInstanceId(instanceId)` recursive or not, does anybody have a use case for this ?

EricKuck:
I think you're right. The `getChildControllerWithInstanceId` method probably shouldn't have been public in the first place, as there isn't a valid use for using this method (that I can think of). The whole concept of an instance id should probably be an implementation detail of the library, not something that's exposed to the developer.

Let's do as you suggested - deprecate the original `getChildControllerWithInstanceId` method and add this new method as a package private method. Thanks for the thoughtful discussion.

adi1133:
I have updated the PR. I suggest squashing the commits using [github](https://help.github.com/articles/about-pull-request-merge-squashing/).

EricKuck:
Looks great, thanks!

PR #41, opened 2016-04-29
Travis - Add permission for deploy_snapshot.sh

sockeqwe:


adi1133:
Dosen't git respect the executable bit of the commited file when checking out?
I don't see people having the same issue with the bash gradle wrapper.

JakeWharton:
Yes

PR #42, opened 2016-04-29
adding missing snapshot auth credentials

sockeqwe:
I have noticed that currently travis can't push snapshots because auth credentials are missing (http status code 401).

I have added them to maven push script.

All you have to do is merge this request and then setup the auth credentials with the following commands:
```
travis encrypt -r bluelinelabs/Conductor "NEXUS_USERNAME=InsertYourUsername" --add
travis encrypt -r bluelinelabs/Conductor "NEXUS_PASSWORD=InsertYourPassword" --add
```


gradle-mvn-push.gradle
```diff
@@ -32,11 +32,21 @@ def getSnapshotRepositoryUrl() {
 }
 
 def getRepositoryUsername() {
-  return hasProperty('NEXUS_USERNAME') ? NEXUS_USERNAME : ""
+  def username = "$System.env.NEXUS_USERNAME"
+  if (username == null) {
```
JakeWharton: This will never be null

gradle-mvn-push.gradle
```diff
@@ -32,11 +32,21 @@ def getSnapshotRepositoryUrl() {
 }
 
 def getRepositoryUsername() {
-  return hasProperty('NEXUS_USERNAME') ? NEXUS_USERNAME : ""
```
JakeWharton: You can use `ORG_GRADLE_PROJECT_NEXUS_USERNAME` to set properties from environment variables and retain the original behavior.

sockeqwe: Ah, didn't know about that, thanks!

So basically, this pulll request can be rejected and @EricKuck only has to set:

````
travis encrypt -r bluelinelabs/Conductor "ORG_GRADLE_PROJECT_NEXUS_USERNAME=InsertYourUsername" --add
travis encrypt -r bluelinelabs/Conductor "ORG_GRADLE_PROJECT_NEXUS_PASSWORD=InsertYourPassword" --add
```

right?

JakeWharton: Yep!

PR #51, opened 2016-05-15
Add proguard rules

adi1133:
This will add proguard rules into the generated AAR files to keep the Controller constructor that is used via reflection. 
Are there any other places where reflection may break proguard ?


EricKuck:
Looks good! Can you add a rule for the default constructor too?

adi1133:
I don't think that the default constructor needs a rule because it is never called using reflection.

EricKuck:
It actually is called if the Bundle constructor isn't provided by the
developer. Either one can be used, depending on whether or not the
controller needs args.

adi1133:
Done

PR #60, opened 2016-06-05
Target Controller large bitmap issue fix

shchurov:
`ImageView.setImageURI()` doesn't resize Bitmaps, it may lead to a silent error in logs when you try to pick a high resolution image:
`Bitmap too large to be uploaded into a texture (4160x3120, max=4096x4096)`

One approach is to do it manually, the other is to use a third party library. Since Demo uses Butterknife, Picasso would be a consistent choice.


EricKuck:
Looks good. Thanks.

PR #62, opened 2016-06-10
Added cause to the instantiation exception

shchurov:
It's useful to see the real stacktrace of an exception. Also `e.getMessage()` can be null if it's just a wrapping exception.
As an example:
I had a bug in reading `Parcelable` from controller arguments, it was some sort of casting exception. Android wrapped it with another exception that didn't have any message. As a result in the logs I saw a stacktrace that was pointing to the line 95 of `Controller.java` and a message "An exception occurred while creating a new instance of  YourController. null"


EricKuck:
Thanks!

PR #83, opened 2016-06-30
Fixed typo in readme

shchurov:
There's no `getChildController()` method in `Controller`. I would assume it's a typo.


EricKuck:
Thanks!

PR #84, opened 2016-06-30
[Demo] ParentController crash when pressing back while loading

adi1133:
I am not sure how to fix this without introducing too much complexity but here is a PR anyway.
Possible deal breakers
- usage of RxJava and RxAndroid
- no longer uses removeChild()


EricKuck:
Thanks for the fix! While this is probably closer to what I'd do in a real app, introducing Rx into the demo app might be a little much for some people. The goal is to keep the demo app pretty vanilla so just about anyone can pick it up. I ended up fixing it a different way (basically blocking the back button until the animation is done), since that's how it originally worked before I introduced the bug in the first place.

adi1133:
Good choice :) thanks for the fix.

PR #88, opened 2016-07-05
Fix IndexOutOfBoundsException in ParentController.

robxyy:


EricKuck:
Thanks for the fix! I ended up fixing it a totally different way (essentially blocking the back button), since that's how it was originally worked before i introduced the bug.

PR #89, opened 2016-07-06
Fix VerticalChangeHandler

3flex:
Ensures the previous view is animated on push, and next view is animated on pop.

In my app HorizontalChangeHandler was working fine, but when I switched to VerticalChangeHandler (making no other changes) the animation wouldn't run.

This corrects the issue.


EricKuck:
Thanks for the PR, but this was working as intended. Both controllers aren't meant to animate. The controller being pushed is supposed to animate over the old controller. On pop, it's supposed to drop down out of view. If you want a transition that animates both you're free to add that to your project.

3flex:
Thanks @EricKuck. In that case, should HorizontalChangeHandler be updated to change its behaviour? It currently animates both views (that's where the new code in this PR was adapted from).

EricKuck:
They're both working as designed. These two transitions were what we always used to use before material design and shared element transitions. They're also there to be references, so you don't have to use any of the built-in change handlers.

PR #90, opened 2016-07-08
Added oss sonatype snapshot url

sockeqwe:


EricKuck:
Wow, that was a glaring oversight. Could have sworn that was there before. Thanks!

PR #93, opened 2016-07-14
Document a return value of Router.handleBack()

dimsuz:
I found myself guessing whether or not this `Router.handleBack()` function will return what I expect it to return, had to dig in sources, so I decided it would help to add a bit of documentation :)

(not sure if targeting `develop` branch is a right way, please correct if not)


EricKuck:
Nice addition. Can you capitalize the first letter though?

dimsuz:
Oh, right, didn't notice it was out of existing style. Fixed.

EricKuck:
Thanks!

PR #96, opened 2016-07-15
Nitpicks

adi1133:
Some minor cleanups.


demo/src/main/java/com/bluelinelabs/conductor/demo/MainActivity.java
```diff
@@ -12,22 +12,20 @@
...
     private Router router;
-    private Unbinder unbinder;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         setContentView(R.layout.activity_main);
-        unbinder = ButterKnife.bind(this);
+        ButterKnife.bind(this);
```
EricKuck: Why remove the unbinder?

adi1133: All references to the activity (and its views) are released after onDestory(). The unbinder is not required.

dimsuz: Imho, it doesn't hurt to release them manually, also shows off this ButterKnife feature (which is a bit new if I'm right)
Also helps to get used to using it in Controllers... Maybe a comment about not-being-required here would help. 

On the other hand it is also shown off in ButterKnifeController... :)

adi1133: In ButterKnifeController it is required to not leak memory.

PR #104, opened 2016-08-24
Remove listeners after AnimatorChangeHandler animator ends/is canceled

### LINKED ISSUE #103: AnimatorChangeHandler leaks destroyed controllers
It looks like `AnimatorChangeHandler` leaks destroyed controllers through its `Animator`'s listener.

[Example](https://github.com/hannesstruss/Conductor/tree/hs/refwatch-demo/demo/src/main/java/de/hannesstruss/refwatchdemo) (launch, click "Child 2" and wait for LeakCanary notification)
[LeakCanary trace](https://gist.github.com/hannesstruss/287cf2ffdf0945508a271209259fa0f3)

The fix seems to be pretty straightforward, will open a PR in a bit.

hannesstruss:
Fixes #103


EricKuck:
Thanks, good catch!

PR #154, opened 2016-11-07
added first level of deep linking to demo-app

tmtron:
with this change, you can start controllers via a deep link directly by using the ID of their corresponding  `HomeDemoModel` enumeration.

Example: 
> adb shell am start -W -a android.intent.action.VIEW -d "http://bluelinelabs.com/demo/NAVIGATION" com.bluelinelabs.conductor.demo

note: this change only handles the first level.  
there could also be multiple levels of deep linking: e.g. link to the 2nd child-controller in the master-detail controller using this URL: `http://bluelinelabs.com/demo/MASTER_DETAIL?item=1`


demo/src/main/java/com/bluelinelabs/conductor/demo/MainActivity.java
```diff
@@ -31,7 +36,49 @@ protected void onCreate(Bundle savedInstanceState) {
 
         router = Conductor.attachRouter(this, container, savedInstanceState);
         if (!router.hasRootController()) {
-            router.setRoot(RouterTransaction.with(new HomeController()));
+
+            HomeController homeController = new HomeController();
+            router.setRoot(RouterTransaction.with(homeController));
+            handleIntentDataUri(homeController);
```
EricKuck: Won't deep linking not work if you've already launched the app if you use this strategy?

kiptoomm:
@PaulWoitaschek @tmtron @EricKuck any plan to release this or its alternatives any time soon?

PR #167, opened 2016-11-12
fixed misleading log messages in RxLifecycleController.onViewBound()

tmtron:


EricKuck:
This was actually intentional. I wanted to log messages to map to lifecycle events in the Conductor library, not abstracted methods from the demo app.

PR #189, opened 2016-12-27
Recreate views on config changes

hansonchris:
This is typically necessary when the Activity has specified in the manifest that it will handle configuration changes such as an orientation change.

If the top Controller.setRecreateViewOnConfigChange(true) has been called before a configuration change happens, that Controller's current view will be removed and a recreated view will be added to the parent.

For Controllers that are lower in the stack, those Controllers' view will only be recreated if the Controller also has a RetainViewMode of RetainViewMode#RETAIN_DETACH.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -86,6 +86,7 @@ public abstract class Controller {
     private final ArrayList<String> requestedPermissions = new ArrayList<>();
     private final ArrayList<RouterRequiringFunc> onRouterSetListeners = new ArrayList<>();
     private WeakReference<View> destroyedView;
+    private boolean recreateViewOnConfigChange;
 
     @NonNull
     static Controller newInstance(@NonNull Bundle bundle) {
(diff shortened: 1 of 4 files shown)


hansonchris:
@EricKuck would you mind taking a look at this when you have a chance?

hansonchris:
Rebased, resolved conflicts, was bee8128f85af6aa5823549939c7a04576aea989f

hansonchris:
Rebased, resolved conflicts, was 90b7add15305d54409a14b512ff51366ba65fa90

EricKuck:
Recreating views on _all_ config changes seems a little heavy-handed, doesn't it? I can see people potentially wanting to do this for landscape -> portrait, but what if they don't want to recreate views for keyboard availability events or multi-window changes?

Tagakov:
Maybe we can allow user to pass predicate which would filter appropriate configchanges for view hierarchy recreation? Approach with view hierarchy recreation without destroying activity seems very usable

Tagakov:
Or maybe it would be better to just add recreateViews() method to router allowing users to invoke it in Activity::onConfigurationChanged?

Tagakov:
ping?

CebotariM:
Any updates on this?

PR #216, opened 2017-01-31
Fixes issue when retrieving an existing controller from a ControllerPagerAdapter

blazeroni:
If you page away from a controller then return, `ControllerPagerAdapter#getController()` will not return the controller.  This is due to the controller being removed from `visiblePageIds` in `destroyItem()`, but never added back when the controller is created again.


EricKuck:
Great find. Thanks!

PR #225, opened 2017-02-08
List simplification

PaulWoitaschek:
* Singleton list for better performance
* Use builtin methods


EricKuck:
Great, thanks! Got so used to Kotlin's `listOf` that I forgot about Java's `singletonList`.

PaulWoitaschek:
I just did the same in my last project a few moth ago when I discovered that there is already an `Arrays.asList` built into java.
But yes, I love how kotlin makes you lazy by making you at the same time more performant ;-)

Issue #214, opened 2017-01-27
TransitionChangeHandler


nomisRev:
`java
 @Override
 public final boolean removesFromViewOnPush() {
    return true;
}
`

I'm doing a shared element transition of a fab to a dialog. similar as in Plaid. But since I'm transition to an overlay, I want the `from` view to not be removed. But `removesFromViewOnPush` is final and true in TransitionChangeHandler opposed of non-final in AnimationChangeHandler.

Any particular reason why it's final here? Or how I could achieve this with the current API?

EricKuck:
If the `from` view isn't removed (or set to `GONE`), how does the shared element transition work? Do you have a link to something similar?

nomisRev:
That is what I currently have. And it is working as "expected". You start `        TransitionManager.beginDelayedTransition(container, transition)` and then it will check what transition to do depending on what changes. So traditionally you do `remove(from)` and `add(to)`, and that will cause a transition `from -> to`

In my case I want to only do a transition on the FAB, the weird issue I'm seeing now is that the shared element works but the `to` ViewController jumps behind the `from` ViewController after `onChangeCompleted`. If I figure that out, I think I can get it working.

(I have a fairly simply layout list + FAB, when the FAB is clicked it transitions to the middle of the screen in a new VC). If I have that working I can add a custom ChangeBounds which will do a shared element with change bounds from FAB -> ViewGroup.


```java
    override fun performChange(container: ViewGroup, from: View?, to: View?, isPush: Boolean, changeListener: ControllerChangeHandler.ControllerChangeCompletedListener) {
        if (canceled) {
            changeListener.onChangeCompleted()
            return
        }

        val transition = getTransition(container, from, to, isPush)
        transition.addListener(object : Transition.TransitionListener {
            override fun onTransitionStart(transition: Transition) {}

            override fun onTransitionEnd(transition: Transition) {
                changeListener.onChangeCompleted()
            }

            override fun onTransitionCancel(transition: Transition) {
                changeListener.onChangeCompleted()
            }

            override fun onTransitionPause(transition: Transition) {}

            override fun onTransitionResume(transition: Transition) {}
        })

        TransitionManager.beginDelayedTransition(container, transition)
        if (from != null) {
            val view = from.find<FloatingActionButton>(R.id.fab)
            (from as ViewGroup).removeView(view)
//            container.removeView(from)
        }
        if (to != null && to.parent == null) {
            container.addView(to)
        }
    }
```

That is my push CH.

[...]

PR #229, opened 2017-02-10
DialogToFabTransition

nomisRev:
I've added Dialog to fab transition in demo which solves issue #214 but it brings an additional issue with transition when working view based (Unrelated to Conductor).

I've added some comments in com.bluelinelabs.conductor.demo.changehandler.FabToDialogTransitionChangeHandler describing the problem. And I also by passed the problem by using a different pop transition (dialog -> fab)

 What I have not included here is a way to communicate between the dialog and the host view controller. In my original setup I do this by injecting a RxRelay in both. This could probably be in various different ways but I thought this way irrelevant here.


EricKuck:
This looks great. Do you have any idea why the transition from the Dialog back to the FAB looks so stretched out? I'd like to fix that before merging this in. I'll be looking at it too, just wondering if you had any ideas.

EricKuck:
Just read your comments about why the pop transition works the way it does, so I guess you don't have to answer that question. I'd still like to try to get to the bottom of it though! Earlier you mentioned you'd done some things with the z-order to make things work better. What had you done then?

nomisRev:
@EricKuck I'd really like to get to the bottom of it but I've spent quite some time trying to figure it out already and I needed some time off before looking into it again. There are a couple of things you could try but nothing will really fix the issue. Because even if you manage to fix the ghost state of the ImageButton (fab) the transition will still look quite off compared the transition in Plaid.

You can compare the transition when commenting the `if else` in the `getTransition` method. Only real issue I found there is when you inspect the View Hierarchy the ImageButton is still holding onto a Overlay after the transition. So my guess is that the difference in pop transition is caused by the same issue that creates the ghost state of the fab. So the Overlay is currently the only thing I can think of that is causing all of this, Activity transitions use overlay quite differently than what is happening here. But I haven't really took the time to explore that path as well, and all cases I can think of are quite dirty hacks. (i.e. removing and re-inflating the fab with every transition, while that probably will not result in any performance issues it will result in very ugly code.)

I also tried the `io.plaidapp.ui.transitions.MorphTransform` that is used in the original pop transition `io.plaidapp.ui.transitions.MorphDialogToFab` but it looks almost identical to `ChangeBounds` with the custom GravityArcMotion. Since it seems to be intended for transition between a rect shaped view and a dialog.

The z-order thing was totally unrelated to the transition, but was something I ran into when trying to show the new ViewController on top of the current one. Since it was the first time I really looked into doing advanced transitions with a View based architecture I ran into a lot of bumps along the road before achieving what I currently have.

EricKuck:
I think I actually got it working like Plaid. Took FOREVER and your base was super helpful. I'll do a PR on your repo so you can look over what I did.

EricKuck:
Actually now that I think about this more - what do you think about putting the FAB on the home screen and making it take the user to the about dialog (restyled to look like the new one)? That way it wouldn't be hidden behind a row in the home controller? I guess my only concern with that is that there's already lots of dots on the home controller, which could make that confusing.

nomisRev:
I guess that really depends on what you want to do with the demo. If you just want to add some of Plaid transitions, or want to recreate Plaid all together. Recreating Plaid however will probably be quite time consuming.

However, I'm not quite sure were all of this would fit into. I've come across quite some issues with questions like a demo with Dagger 2 or the use of "the Navigator pattern" I mentioned somewhere. I'm also still looking for a better solution of handling the toolbar and turning on and off shadows on the toolbar myself.

Maybe a more advanced sample along side the demo to showcase patterns like that. Which could showcase a production stable codebase compared to showcasing Conductor's features in a demo. But then again draw the line where. Should you include Rx, Kotlin, Dagger 2, ... ?

IMHO, a FAB in the right corner with a scroll behavior will not be confusing since the size would be different, it would have elevation and an icon. Just let me know and I'll change it :)

PR #248, opened 2017-03-16
DialogController like DialogFragment

dmdevgo:
Implemented DialogController as a wrapper over standard android dialog.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
@@ -1,7 +1,6 @@
 package com.bluelinelabs.conductor;
 
 import android.app.Dialog;
-import android.content.Context;
 import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
@@ -16,9 +15,9 @@ import com.bluelinelabs.conductor.changehandler.FadeChangeHandler;
  * A controller that displays a dialog window, floating on top of its activity's window.
  * This is a wrapper over {@link Dialog} object like {@link android.app.DialogFragment}.
  *
- * <p>Implementations should override this class and implement {@link #onCreateDialog(Context context)} to create a custom dialog, such as an {@link android.app.AlertDialog}
+ * <p>Implementations should override this class and implement {@link #onCreateDialog(Bundle)} to create a custom dialog, such as an {@link android.app.AlertDialog}
  */
-public abstract class DialogController extends Controller {
+public abstract class DialogController extends RestoreViewOnCreateController {
 
     private static final String SAVED_DIALOG_STATE_TAG = "android:savedDialogState";
 
@@ -31,8 +30,9 @@ public abstract class DialogController extends Controller {
 
     @NonNull
     @Override
-    final protected View onCreateView(@NonNull LayoutInflater inflater, @NonNull ViewGroup container) {
-        dialog = onCreateDialog(getActivity());
+    final protected View onCreateView(@NonNull LayoutInflater inflater, @NonNull ViewGroup container, @Nullable Bundle savedViewState) {
+        dialog = onCreateDialog(savedViewState);
+        //noinspection ConstantConditions
         dialog.setOwnerActivity(getActivity());
         dialog.setOnDismissListener(new DialogInterface.OnDismissListener() {
             @Override
@@ -40,16 +40,13 @@ public abstract class DialogController extends Controller {
                 dismissDialog();
             }
         });
-        return new View(getActivity());//stub view
-    }
-
-    @Override
-    protected void onRestoreViewState(@NonNull View view, @NonNull Bundle savedViewState) {
-        super.onRestoreViewState(view, savedViewState);
-        Bundle dialogState = savedViewState.getBundle(SAVED_DIALOG_STATE_TAG);
-        if (dialogState != null) {
-            dialog.onRestoreInstanceState(dialogState);
+        if (savedViewState != null) {
+            Bundle dialogState = savedViewState.getBundle(SAVED_DIALOG_STATE_TAG);
+            if (dialogState != null) {
+                dialog.onRestoreInstanceState(dialogState);
+            }
         }
+        return new View(getActivity());//stub view
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


hannesstruss:
I haven't looked at the source closely, but could this be just a separate library?

conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
```diff
@@ -0,0 +1,97 @@
...
+        router.pushController(RouterTransaction.with(this)
+                .pushChangeHandler(new FadeChangeHandler(false))
+                .popChangeHandler(new FadeChangeHandler()));
+    }
+
+    public void dismissDialog() {
+        if (mDismissed) {
+            return;
+        }
+        getRouter().popCurrentController();
```
PaulWoitaschek: What if there are multiple dialogs. It would make sense to let the user supply a tag.

conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
```diff
@@ -0,0 +1,97 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import com.bluelinelabs.conductor.changehandler.FadeChangeHandler;
+
+/**
+ * @author Dmitriy Gorbunov
```
PaulWoitaschek: Remove header and add documentation

demo/src/main/java/com/bluelinelabs/conductor/demo/controllers/AlertDialogController.java
```diff
@@ -0,0 +1,56 @@
...
+import android.support.annotation.NonNull;
+import android.support.v7.app.AlertDialog;
+import android.text.method.LinkMovementMethod;
+import android.widget.TextView;
+
+import com.bluelinelabs.conductor.DialogController;
+import com.bluelinelabs.conductor.demo.util.BundleBuilder;
+
+/**
+ * @author Dmitriy Gorbunov
```
PaulWoitaschek: Remove author

conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+public abstract class DialogController extends Controller {
+
+    private static final String SAVED_DIALOG_STATE_TAG = "android:savedDialogState";
+
+    private Dialog mDialog;
```
PaulWoitaschek: No Hungarian style

demo/src/main/java/com/bluelinelabs/conductor/demo/controllers/AlertDialogController.java
```diff
@@ -0,0 +1,56 @@
...
+                .setMessage(getArgs().getCharSequence(KEY_DESCRIPTION))
+                .setPositiveButton(android.R.string.ok, null)
+                .create();
+
+        alertDialog.setOnShowListener(new DialogInterface.OnShowListener() {
+            @Override
+            public void onShow(DialogInterface dialog) {
+                TextView messageTextView = ((TextView) alertDialog.findViewById(android.R.id.message));
+                if (messageTextView != null) {
+                    //Make the textview clickable
```
PaulWoitaschek: Why?

PaulWoitaschek (changes requested):

demo/src/main/java/com/bluelinelabs/conductor/demo/controllers/AlertDialogController.java
dmdevgo: Typo, make links clickable in textview

conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Display the dialog, create a transaction and pushing the controller.
+     * @param router The router on which the transaction will be applied
+     * @param tag The tag for this controller
+     */
+    public void showDialog(@NonNull Router router, @Nullable String tag) {
+        dismissed = false;
+        router.pushController(RouterTransaction.with(this)
+                .pushChangeHandler(new FadeChangeHandler(false))
```
mradzinski: Wouldn't it be better to let the implementer supply a ```ChangeHandler``` instead of forcing one? I guess the ```showDialog``` method could be overloaded to allow this.

PaulWoitaschek: I think so too. The dialog needs to call `removesFromViewOnPush` on the push change handler in that case and throw an exception if it's true.
I also ask myself if it's possible to skip the whole showDialog stuff and use the controller lifecycle callbacks to handle the dialog.

conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
```diff
@@ -0,0 +1,116 @@
...
+ * <p>Implementations should override this class and implement {@link #onCreateDialog(Bundle)} to create a custom dialog, such as an {@link android.app.AlertDialog}
+ */
+public abstract class DialogController extends RestoreViewOnCreateController {
+
+    private static final String SAVED_DIALOG_STATE_TAG = "android:savedDialogState";
+
+    private Dialog dialog;
+    private boolean dismissed;
+
+    public DialogController(@Nullable Bundle args) {
```
PaulWoitaschek: Please add a no-args constructor too

conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
dmdevgo: The `DialogController` return a stub view because a method `onCreateView` requires non-nullable result. This view not visible for user. At the same time we should not remove the view of controller on top of which the dialog is displayed. Therefore, there are no views that need to be animated. I think apply `SimpleSwapChangeHandler(false)` instead `FadeChangeHandler(false)`.

esafirm:
Any update on this guys?

ghost:
Why use stub view?
This is not obvious that we will get just a stub when using `Controller.getView()`

There is must be a way to pass `Dialog`'s view as `Controller`'s view.
This is very common when we want to implement custom view for the `Dialog`.

Now we forced to use `getDialog().findViewById()` instead of just `getView()`

PaulWoitaschek:
Thank you for your efford! Closing this for the same reason as #467

PR #249, opened 2017-03-18
Added a ChangeHandler using Transitions-Everywhere transitions

zawadz88:
Hi,
Thanks for the great library!
I was going through the samples and when I stumbled upon the Transitions sample I thought it might work with transitions from the Support library or from Transitions-Everywhere library.

Hence, I've added a ChangeHandler using Transitions-Everywhere transitions instead of native Transitions to support pre-Lollipop devices as Support transitions are more limited (no support for ArcMotion or ChangeClipBounds yet).

I thought about creating a separate project for this, however I figured it might be more useful if it's provided out-of-the-box with this library.

Let me know what do you think!


EricKuck:
I think this is awesome! However, it seems like maybe this functionality is tangential to the purpose of this library and should probably be provided as a separate library. If you want to create a repo for this and push it to maven, I'll gladly link to it from the main readme.

zawadz88:
Hi @EricKuck,
It took me some time to actually sit down and do it, but here it is: https://github.com/zawadz88/Conductor-TransitionsEverywhere
Thanks again for a great library!

PR #250, opened 2017-03-21
Fixed readme table

Namolem:
The second table's header is broken


EricKuck:
Ah, good catch. Thanks!

PR #251, opened 2017-03-21
Fix readme image display

Namolem:
Github doesn't understand spaces in file name


EricKuck:
Man, they're killing me with these new markdown changes. Thanks!

PR #258, opened 2017-04-04
Pass along the correct view state

fergusonm:
The bundle passed in onCreateView is not the one created in onSaveViewState but the "enclosing" view state.


EricKuck:
Great catch, thanks!

PR #283, opened 2017-05-02
Add convenience method Controller.getContext()

cachapa:
Reduces the effort when migrating from Fragments

I recently migrated a project which was heavily using fragments, and it was a (in my opinion unnecessary) pain to go through all of my fragments and replacing getContext calls with getActivity.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -282,6 +282,14 @@ public abstract class Controller {
     }
 
     /**
+     * Convenience method equivalent to calling {@link #getActivity()}
+     */
+    @Nullable
+    public final Context getContext() {
+        return getActivity();
+    }
+
+    /**
      * Returns the Resources from the host Activity or {@code null} if this Controller has not
      * yet been attached to an Activity or if the Activity has been destroyed.
      */


sockeqwe:
Imho this is not an API improvement of a controller. In fact, this will add more confusion as now you can call `getActivity()` or `getContext()`. 

Regarding migrating: with Android Studio you can do find / replace in path which does renaming from getContext() to getActivity() very easy and convenient.
Otherwise, one could create its own controller base class providing a `getContext()` method delegating to `getActivity()` and then use this base controller class instead of plain Controller class. 

Adding getContext () makes Controller's API unnecessary verbose  for sake of "easier migrating from fragments".

cachapa:
Blindly performing find/replace on your code can be dangerous, especially if you're replacing getContext with getActivity since in some cases you can introduce subtle bugs (e.g. getContext on a themed fragment is different from getActivity).

I understand the concern with making the API unnecessarily verbose, but I think in this case it's justified by the fact that replacing fragments is one of this library's main use cases.

PaulWoitaschek:
Its great to keep the API clean. Why create unnecessary legacy code? Add it as a base class if you really need it. Every method less makes the API easier to understand. Especially if it's a duplicate method.

cachapa:
I don't think there's much confusion between getActivity and getContext. The problem doesn't exist in fragments either since you get the object you need for your use (say, layoutInflater requires a context object whereas you can only call finish() on an Activity).

It's actually specialised knowledge that an Activity is a Context object. Granted, every Android dev knows that, but it's not clear just looking at the API.

cachapa:
@PaulWoitaschek I'm fine with going over the code and replacing getContext with getActivity calls.
As I said, I understand the concerns even if in this case I think it would be justified in making Controllers (more of) a drop-in replacement for Fragments.

EricKuck:
While I appreciate the idea of reducing friction when transitioning from Fragments to Controllers, a goal of this library needs to be having a clear API, not being a drop-in replacement to what people were using before. As others have mentioned, `getActivity()`and `getApplicationContext()` already exist for the use of getting a context. If this is useful to some, by all means make a `BaseController` that includes a `getContext()` method!

Issue #260, opened 2017-04-04
[Question] Bottom navigation handling 


EmkaMk:
Hi, 
I am just wondering what is your opinion on handling bottom navigation with Conductor? 
I am aware that I should have one container ViewGroup which would, per say hold the bottom navigation, but each tab click should open some sub screen that would overlay the bottom navigation. 
I saw the demo example for MultipleChildRouters, but I notice that clicking on the next button replaces the parent container altogether with the navigation itself, instead of just filling the sub screen with the new Controller's view. 
Is it possible to have multiple childRouters with the same root ( that being the navigation container )? If it is, who will handle the navigation clicks? Parent Controller? 
 Thanks!

mradzinski:
Given three controllers named MainController, HomeChildController and ProfileChildController, I would architect the navigation in the following way:

MainController has both, a container view (FrameLayout for example) and the bottom navigation view. I would get the child router of MainController using ```getChildRouter(container)```  and attach HomeChildController to that child router root controller (you can do this on ```onCreateView``` for example of the MainController). The rest is just invoking the child router pop and push methods from ```setOnNavigationItemSelectedListener```, so, if I wanted to show the ProfileChildController I would push it, and if I wanted to go back to the HomeChildController I would just pop to root.

EmkaMk:
Yes, but what happens if i have let's say 5 items in the navigation tab? Pop to root doesn't make sense, since I would always get back to the same screen no matter what I chose. I can use pop to tag, but that clears up the stack up to the Controller with the given tag, so I would be recreating the views. This is not something I want to achieve.

mradzinski:
Can you describe the navigation behavior you are trying to achieve? I don't think I'm getting it right

EmkaMk:
Take a look click[https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B3321sZLoP_HQWtkdUY4U3lla0U/components_bottomnavigation_behavior_hierarchy.webm](url)

EmkaMk:
Take a look [click](https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B3321sZLoP_HQWtkdUY4U3lla0U/components_bottomnavigation_behavior_hierarchy.webm)

mradzinski:
[...]

Issue #156, opened 2016-11-07
[QUESTION] How to "emulate ViewFlipper" with conductor?


StephanSchuster:
I have a scenario much like your master/detail sample in landscape mode:

I have a "MainController" whose view has a group of 5 radio buttons and a child container into which the according child controller should be rendered upon button click. But in this special case I want neighter the child controllers nor their views to be destroyed when I change the active child controller (like ViewFlipper). How can I achieve this with conductor? Or is this something that should just not be done???

My current solution:
As far as I understand, a controller which is not in the backstack of a router always gets destroyed on pop - regardless of  "setRetainViewMode()". Right? So first I need to get all of the 5 child controllers to the router attached to the child container of my "MainController". Currently I do this by initially calling getChildRouter(...).pushController(...) for each of my 5 child controllers. Upon button click I call getChildRouter(...).getBackstack(), change the result list to have the new active controller on the end, and call getChildRouter(...).setBackstack(<modified list>). Furthermore I need to ensure that none of my 5 child controllers is popped when clicking the native back button. So I override MainController.handleBack() and delegate this to <active child controller>.handleBack(). As a last step I set RetainViewMode.RETAIN_DETACH for each of my 5 child controllers. Now on button click there is only attach/detach called for each child controller. So it seems to work. But is there a better way to achieve this behaviour???

By the way: when I initialize the MainController with getChildRouter(...).setBackstack(5 x child controller) instead of getChildRouter(...).pushController(...) for each of my 5 child controllers, the internal "childBackstack" variable in Controller.java seems to be wrong. It just contains the last entry of the backstack list, not the full list. This seems like a bug to me. However, this is not really relevant for my use case because I override handleBack() anyway.

stephanschuster:
I wrapped my current solution into an abstract base class:

```
import android.os.Bundle;
import android.support.annotation.NonNull;
import android.text.TextUtils;

import com.annimon.stream.Stream;
import com.bluelinelabs.conductor.Controller;
import com.bluelinelabs.conductor.ControllerChangeHandler;
import com.bluelinelabs.conductor.Router;
import com.bluelinelabs.conductor.RouterTransaction;

import java.util.List;

public abstract class SwitchController extends BaseController {

    private Router mChildRouter;

    protected SwitchController() {
    }

    protected SwitchController(Bundle args) {
        super(args);
    }

    protected void init(@NonNull Router childRouter, @NonNull Controller... childControllers) {
[...]

Issue #27, opened 2016-04-21
[Feature] Nested backstacks


mheras:
Hi!

Is there any way to push a [child] controller from within another child controller?

Let's say I'm displaying a child controller (named **A**) in a container view (named **V**), that it is just a part of the whole screen.

What I want is to push another controller from **A** and make its view be contained in **V**.

From what I understand, if we push a controller using the router, that would replace the view of the parent controller, and that's not what I want to achieve.

Thank you in advance!

**EDIT**

My use case is:
- I want a tabbed controller so I'm using `ControllerPagerAdapter`.
- I want its child controllers to have their own backstacks so I can navigate within that "section" and keep the tabbar always visible.

EricKuck:
This is currently unsupported. I'm not sure I really want to support nested backstacks, due to the greatly increased complexity of dealing with such things, both from a library maintainer's point of view and an application developer's point of view.

That being said, your use case doesn't seem unreasonable, so it would be nice to come up with a workable solution. One thing I'm not totally sure about though - would this even work with a ViewPager in the first place? Can you replace one of its pages and expect it to continue working correctly?

sockeqwe:
Im also facing a similar use case: i.e. an E-mail app on a tablet
displaying the inbox (list of mails). When I click on a mail item from
inbox, I would like to show the details to that email. It would be nice if
I could add a child controller to the parents layout dynamically to switch
to a master-detal view with changehandler runing animations. I e.
LayoutTransition or TransitionManager (api19)

Eric Kuck <notifications@github.com> schrieb am Sa., 23. Apr. 2016, 00:54:

EricKuck:
@sockeqwe, I think your use case is a bit different from @mheras's. You're looking for master/detail view, while he's looking for having one backstack per page in a ViewPager. My original thought was that master/detail view applications would use a single Activity with multiple routers. So instead of

```
--------------------------------------------
| Parent Controller                        |
|  --------------------------------------  |
|  |    Child    |    Child             |  |
|  |  Controller |  Controller          |  |
|  |      1      |      2               |  |
|  --------------------------------------  |
--------------------------------------------
```

you'd have

```
--------------------------------------------
| Activity                                 |
|  --------------------------------------  |
|  |    Router    |    Router           |  |
|  |       1      |      2              |  |
|  |      or      |                     |  |
|  |   ViewGroup  |                     |  |
|  --------------------------------------  |
--------------------------------------------
```

[...]

PR #316, opened 2017-06-19
Adds BottomNavigationController to the demo app

chris6647:
Googling around revealed quite a few issues for Conductor showing interrest in Conductor and Bottom Navigation together. This is an example implementation of that, but where each MenuItem has a separate backstack.

The first implementation I did, was something that @EricKuck mentioned in a related issue (https://github.com/bluelinelabs/Conductor/issues/156#issuecomment-259991243) where I created a ChangeHandlerFrameLayout pr MenuItem, but the main problem with that, was how each of the Controllers in each of the ChangeHandlerFrameLayout with their own Router, all had their view active and themselves being in attached state at the same time. We needed to be able to trust the lifecycle methods, and thus we gave this another attempt.

After the multiple ViewGroup/Router attempt, I tried managing the backstack state by keeping track of the RouterTransactions, but the Bundle constructor was package private, and thus I couldn't easily recreate it. Same goes internally used Backstack implementation. 

So even if there was overhead in saving/restoring the entire Router pr MenuItem, it allowed me to consistently keep track of them individually and have the handleBack() and restoration functionality as we wanted. Which is what this implementation is.

I hope it will help someone - even if this doesn't make it into the demo app, and perhaps get some comments to improve the solution or do something completely different all together.

Some relevant issues:
https://github.com/bluelinelabs/Conductor/issues/27#issuecomment-224061641
https://github.com/bluelinelabs/Conductor/issues/260


StefMa:
I really love it and it works perfect.
Thanks @chris6647 

I do really like to see it into the conductor samples.
Any reasons why not still merged @EricKuck ? 🤔

chris6647:
@StefMa Thanks for your kind words! Please do check out my recent changes to it, as it improves it quite a bit, and fixes some bugs :)

PR #319, opened 2017-06-25
Fixes ControllerScopeProvider.from return type

hannesstruss:
I might be missing something, but looks like the `.from` method should return the specific type instead of the (raw) super type.

Also hope the newline at EOF is okay, if not, happy to remove it ;)


EricKuck:
I just left the type off because that's what Uber did in their examples. But yeah, might as well return what it is I suppose. Thanks!

PR #328, opened 2017-07-18
Expose parent constructor with Bundle as parameter

luizgrp:
Expose constructor of parent class with Bundle as parameter in LifecycleController.


EricKuck:
Great catch. I don't think there's a reason to move `lifecycleRegistryOwner`'s initialization into the constructor though. Can you move that back to where it was and just call `super` from both constructors?

luizgrp:
I agree, done 👍 
To be honest I had just followed a "pattern" I saw [here](https://github.com/bluelinelabs/Conductor/blob/09df937c8825bffefba935480df8ce5780dc1737/conductor-modules/rxlifecycle2/src/main/java/com/bluelinelabs/conductor/rxlifecycle2/RxController.java) (initialization of the class variable into the constructor and constructor without parameters calling the other one).

luizgrp:
By the way, thanks for this amazing library!

EricKuck:
Haha I didn't notice I did it the other way in the lifecycle module. I'll change that. Thanks for the contribution!

PaulWoitaschek:
It's now inconsistent with `LifecycleRestoreViewOnCreateController`

PR #335, opened 2017-07-27
Only show options menu of the active item in a ViewPager

### LINKED ISSUE #244: ActionBar items visible from offscreen controllers with RouterPagerAdapter
I'm using the `RouterPagerAdapter` class from `conductor-support` and noticing that when child controllers define their own options menu (e.g. `setHasOptionsMenu(true)`), that their option menu items are visible when they're either left or right of the currently active tab (e.g. when they're not yet visible).

My goal is to have contextual menu items for each child controller, and have the menu items change as you scroll through the pager. Is this currently possible or is it a bug in the library? 

Note that I've tried setting the ViewPager's offscreen page limit to 0, but it had no effect on this.

inorichi:
This PR updates `RouterPagerAdapter` in order to show the toolbar's menu items of the active (primary) page instead of all of them.

It uses the same logic as in [FragmentPagerAdapter](https://github.com/android/platform_frameworks_support/blob/master/fragment/java/android/support/v4/app/FragmentPagerAdapter.java) and I've been using it for a while with no issues, though I'm not sure if the `for` loop should be changed and only enable the options menu of the top controller.

Fixes #244.

### DIFF
--- a/conductor-modules/support/src/main/java/com/bluelinelabs/conductor/support/RouterPagerAdapter.java
+++ b/conductor-modules/support/src/main/java/com/bluelinelabs/conductor/support/RouterPagerAdapter.java
@@ -30,6 +30,7 @@ public abstract class RouterPagerAdapter extends PagerAdapter {
     private SparseArray<Bundle> savedPages = new SparseArray<>();
     private SparseArray<Router> visibleRouters = new SparseArray<>();
     private ArrayList<Integer> savedPageHistory = new ArrayList<>();
+    private Router currentPrimaryRouter;
 
     /**
      * Creates a new RouterPagerAdapter using the passed host.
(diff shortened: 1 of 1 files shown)


EricKuck:
This looks great to me. The only change I'd like to see if setting `primaryRouter` to `null` if it's destroyed in `destroyItem`. This *shouldn't* ever happen, but I've had enough issues with ViewPagers that it seems worth making sure.

inorichi:
`destroyItem` destroys adjacent pages, depending on the specified number in `offscreenPageLimit`, so I think setting `primaryRouter` to `null` there isn't right, unless I'm missing something.

EricKuck:
I just meant a simple 

```
if (object == primaryRouter) {
    primaryRouter = null;
}
```

not clearing it no matter what. But yeah you're right, should never be necessary. I just really don't trust ViewPagers :)

PR #364, opened 2017-10-06
Styles snapshot section

StefMa:
* Add missing curly braces
* Make snapshot bold and remove double point
* Fix typo

bildschirmfoto 2017-10-06 um 09 35 06

I think it looks way better now 🙃


EricKuck:
Thanks!

PR #383, opened 2017-11-19
Improved lifecycle integration of Conductor and Architecture Components

StephanSchuster:
Previous mapping was:
- () -> ON_CREATE -> CREATED
- preCreateView -> ON_START -> STARTED
- preAttach -> ON_RESUME -> RESUMED
- preDetach -> ON_PAUSE -> STARTED
- preDestroyView -> ON_STOP -> CREATED
- preDestroy -> ON_DESTROY -> DESTROYED

New mapping is:
- () -> INITIALIZED
- postContextAvailable -> ON_CREATE -> CREATED
- postCreateView -> ON_START -> STARTED
- postAttach -> ON_RESUME -> RESUMED
- preDetach -> ON_PAUSE -> STARTED
- preDestroyView -> ON_STOP -> CREATED
- preContextUnavailable -> // do nothing
- preDestroy -> ON_DESTROY -> DESTROYED

Change-Id: I9f65480018dc1204dc89a8065fa7c7f554ad7dd2


tomblenz:
This is necessary to update arch components to the 1.0 release as well.  @EricKuck

ghost:
@StephanSchuster  is it related: https://github.com/bluelinelabs/Conductor/issues/395??

miquelbeltran:
I am looking at the comments at https://github.com/bluelinelabs/Conductor/issues/395#issuecomment-364639571 and I think in order to avoid keeping subscriptions to LiveData we need to be able to unsubscribe when the Controller loses its context.

@tomblenz Wouldn't be better to have ON_DESTROY on the preContextUnavailable in order to match ON_CREATE in the postContextAvailable? Instead of in preDestroy. If not, ON_DESTROY will only be called when the Controller is finally destroyed and not on configuration changes. (apologies beforehand if I am not understanding it correctly)

tomblenz:
@miquelbeltran Controllers survive config changes, so I think it would be improper to report destruction.  In my testing all seemed well.  I'm currently busy with other tasks if you want to try it yourself.

miquelbeltran:
@tomblenz the problem I see is that if we observe to LiveData using this LifecycleController (or the one that is currently in the library) the observer won't be removed when the view is destroyed. 

That should be the expected behaviour and Fragments suffer the same problem: https://medium.com/@BladeCoder/architecture-components-pitfalls-part-1-9300dd969808

If we want to use LifecycleController to observe LiveData then we need to be sure that the observer is removed onDestroyView.

So far from my tests the only way is to have this more simplified lifecycle:

```
            @Override
            public void preCreateView(@NonNull Controller controller) {
                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_CREATE);
            }

            @Override
            public void postAttach(@NonNull Controller controller, @NonNull View view) {
                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_START);
                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_RESUME);
            }

            @Override
            public void preDetach(@NonNull Controller controller, @NonNull View view) {
                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_PAUSE);
                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_STOP);
            }

            @Override
            public void postDestroyView(@NonNull Controller controller) {
                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_DESTROY);
            }

```

I have set a small project that shows this: https://github.com/miquelbeltran/conductor-viewmodel

If you run the app using the Lifecycle events from this PR (you will have to modify the code) you will see how after a configuration change the `viewModel.liveData` still has the previous observer.

A way to remove the observer from the LiveData is to have ON_DESTROY in postDestroyView.

conductor-modules/arch-components-lifecycle/src/main/java/com/bluelinelabs/conductor/archlifecycle/ControllerLifecycleOwner.java
```diff
@@ -0,0 +1,66 @@
...
+import android.content.Context;
+import android.support.annotation.NonNull;
+import android.view.View;
+
+import com.bluelinelabs.conductor.Controller;
+import com.bluelinelabs.conductor.Controller.LifecycleListener;
+
+public class ControllerLifecycleOwner implements LifecycleOwner {
+
+    private final LifecycleRegistry mLifecycleRegistry;
```
miquelbeltran, line 16: I'd avoid using Hungarian Notation to keep consistency with the rest of the library.

kittinunf:
I do have a similar need and ended up creating my own version of this. And after a close comparison, it is almost exactly the same with implementation here.

Can we see this PR proceed? How can get this going?

EricKuck (approved):
Thanks for the great PR. I'll take out the hungarian notation after merging, so don't worry about that.

Issue #395, opened 2017-12-14
Compatibility with Android Architecture Components


tomblenz:
I believe that the issues raised in https://github.com/bluelinelabs/Conductor/issues/302 merit addressing.

Due to tight coupling with Fragments, the Arch Component library doesn't play nicely with Conductor.  Controllers must use their Activity as their VM provider, which binds them to the Activity lifecycle.  This implies some serious problems

1. The ViewModel will be retained for the lifetime of the Activity - forever for many Conductor apps that use a single Activity
2. Controllers that are used in multiple places in one Activity all share the same ViewModel.  This is obviously a problem for more generically designed Controllers (eg. a controller that displays a news article) but also means that Controllers that are disposed and re-added later are provided a stale ViewModel.  (A workaround for this is to supply a unique key to the ViewModelProvider)

In addition, the current implementation of the experimental `LifecycleController` does not have parity with Activities/Fragments in that LiveData will continue to emit data to the controller after the underlying Activity is stopped.

I'll repeat what I said in the linked issue because I think it is important,

> I simply consider AC to be so strong that even minor trouble integrating it with other libs should be addressed as a priority. I forsee AC to be a strong part of the future of Android development and it's important to keep up in the formative stages.

I'm not sure how these problems can be solved.

tomblenz:
I've been looking into the internals and I believe this solution may work for 1 and 2 - probably needs some eyes and testing over it.

```
object ConductorViewModelProviders {

    private var defaultFactory: ViewModelProvider.Factory? = null
    
    private fun initializeFactoryIfNeeded(application: Application): ViewModelProvider.Factory {
        return defaultFactory ?: ViewModelProviders.DefaultFactory(application).apply { defaultFactory = this }
    }
    
    private fun checkApplication(controller: Controller): Application {
        return controller.activity?.application ?: throw IllegalStateException("Your controller is not yet attached to Application.")
    }
    
    fun of(controller: LifecycleController): ViewModelProvider {
        return ViewModelProvider(ConductorViewModelStores.of(controller),
                initializeFactoryIfNeeded(checkApplication(controller))
        )
    }
    
    fun of(controller: LifecycleController, factory: ViewModelProvider.Factory): ViewModelProvider {
        checkApplication(controller)
        return ViewModelProvider(ConductorViewModelStores.of(controller), factory)
    }
}

object ConductorViewModelStores {

    private val stores = mutableMapOf<LifecycleController, ViewModelStore>()

[...]

PR #405, opened 2018-01-26
Adds ViewModel support to the Arch. Comp. library

miquelbeltran:
* A ViewModelController that has its own ViewModelStore
* The ViewModelController extends LifecycleController so can be used to subscribe to LiveData
* Also provides a ViewModelProvider factory method
* The Architecture Components library version is updated to 1.1.0
* Includes demo usage in ViewModelDemoController

Based on the discussion in https://github.com/bluelinelabs/Conductor/issues/395

This solution is also available as standalone library: https://github.com/miquelbeltran/conductor-viewmodel/tree/master which also fixes the Lifecycle events to handle the LiveData observer subscription properly.

Currently this PR depends on https://github.com/bluelinelabs/Conductor/pull/383 as well, because the current implementation of the LifecycleController can't be used to observe LiveData as the LiveData observer is not being removed in onDestroyView.


dependencies.gradle
```diff
@@ -43,6 +43,7 @@ ext {
     autodispose = "com.uber.autodispose:autodispose:$autodisposeVersion"
 
     archComopnentsLifecycle = "android.arch.lifecycle:runtime:$archComponentsVersion"
+    archComopnentsExtensions = "android.arch.lifecycle:extensions:$archComponentsVersion"
```
jshvarts: Comopnents -> Components

miquelbeltran: thanks @jshvarts I fixed the typo and removed the runtime import, as the extensions already depend on it so it's not required to explicitly add it

conductor-modules/arch-components-lifecycle/src/main/java/com/bluelinelabs/conductor/archlifecycle/ViewModelController.java
```diff
@@ -0,0 +1,18 @@
...
+
+import android.arch.lifecycle.ViewModelProvider;
+import android.arch.lifecycle.ViewModelProviders;
+import android.arch.lifecycle.ViewModelStore;
+
+public abstract class ViewModelController extends LifecycleController {
+
+    private ViewModelStore viewModelStore = new ViewModelStore();
+
+    protected ViewModelProvider viewModelProvider() {
```
jshvarts: `new ViewModelProviders.DefaultFactory()` is deprecated. Can it be dropped here?

miquelbeltran: Thanks! I did not realize 1.1.0 deprecated the method. It was replaced with https://developer.android.com/reference/android/arch/lifecycle/ViewModelProvider.AndroidViewModelFactory.html

jshvarts:
I just tried running the Arch Components View Model demo and got a crash:

```
FATAL EXCEPTION: main
Process: com.bluelinelabs.conductor.demo, PID: 11519
java.lang.RuntimeException: Cannot create an instance of class com.bluelinelabs.conductor.demo.controllers.viewmodel.ViewModelDemo
at android.arch.lifecycle.ViewModelProvider$NewInstanceFactory.create(ViewModelProvider.java:154)
at android.arch.lifecycle.ViewModelProvider$AndroidViewModelFactory.create(ViewModelProvider.java:208)
at android.arch.lifecycle.ViewModelProvider.get(ViewModelProvider.java:133)
at android.arch.lifecycle.ViewModelProvider.get(ViewModelProvider.java:101)
at com.bluelinelabs.conductor.demo.controllers.viewmodel.ViewModelDemoController.onCreateView(ViewModelDemoController.java:49)
at com.bluelinelabs.conductor.Controller.inflate(Controller.java:986)
at com.bluelinelabs.conductor.ControllerChangeHandler.executeChange(ControllerChangeHandler.java:190)
at com.bluelinelabs.conductor.ControllerChangeHandler.executeChange(ControllerChangeHandler.java:154)
at com.bluelinelabs.conductor.Router.performControllerChange(Router.java:792)
at com.bluelinelabs.conductor.Router.performControllerChange(Router.java:762)
at com.bluelinelabs.conductor.Router.performControllerChange(Router.java:744)
at com.bluelinelabs.conductor.Router.pushController(Router.java:179)
at com.bluelinelabs.conductor.demo.controllers.ExternalModulesController.onModelRowClick(ExternalModulesController.java:97)
at com.bluelinelabs.conductor.demo.controllers.ExternalModulesController$AdditionalModulesAdapter$ViewHolder.onRowClick(ExternalModulesController.java:148)
at com.bluelinelabs.conductor.demo.controllers.ExternalModulesController$AdditionalModulesAdapter$ViewHolder_ViewBinding$1.doClick(ExternalModulesController$AdditionalModulesAdapter$ViewHolder_ViewBinding.java:33)
at butterknife.internal.DebouncingOnClickListener.onClick(DebouncingOnClickListener.java:22)
at android.view.View.performClick(View.java:6205)
at android.view.View$PerformClick.run(View.java:23653)
at android.os.Handler.handleCallback(Handler.java:751)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:154)
at android.app.ActivityThread.main(ActivityThread.java:6682)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1520)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1410)
Caused by: java.lang.IllegalAccessException: java.lang.Class<com.bluelinelabs.conductor.demo.controllers.viewmodel.ViewModelDemo> is not accessible from java.lang.Class<android.arch.lifecycle.ViewModelProvider$NewInstanceFactory>
at java.lang.Class.newInstance(Native Method)
at android.arch.lifecycle.ViewModelProvider$NewInstanceFactory.create(ViewModelProvider.java:150)
at android.arch.lifecycle.ViewModelProvider$AndroidViewModelFactory.create(ViewModelProvider.java:208) 
at android.arch.lifecycle.ViewModelProvider.get(ViewModelProvider.java:133) 
at android.arch.lifecycle.ViewModelProvider.get(ViewModelProvider.java:101) 
at com.bluelinelabs.conductor.demo.controllers.viewmodel.ViewModelDemoController.onCreateView(ViewModelDemoController.java:49) 
at com.bluelinelabs.conductor.Controller.inflate(Controller.java:986) 
at com.bluelinelabs.conductor.ControllerChangeHandler.executeChange(ControllerChangeHandler.java:190) 
at com.bluelinelabs.conductor.ControllerChangeHandler.executeChange(ControllerChangeHandler.java:154) 
at com.bluelinelabs.conductor.Router.performControllerChange(Router.java:792) 
at com.bluelinelabs.conductor.Router.performControllerChange(Router.java:762) 
at com.bluelinelabs.conductor.Router.performControllerChange(Router.java:744) 
at com.bluelinelabs.conductor.Router.pushController(Router.java:179) 
at com.bluelinelabs.conductor.demo.controllers.ExternalModulesController.onModelRowClick(ExternalModulesController.java:97) 
at com.bluelinelabs.conductor.demo.controllers.ExternalModulesController$AdditionalModulesAdapter$ViewHolder.onRowClick(ExternalModulesController.java:148) 
at com.bluelinelabs.conductor.demo.controllers.ExternalModulesController$AdditionalModulesAdapter$ViewHolder_ViewBinding$1.doClick(ExternalModulesController$AdditionalModulesAdapter$ViewHolder_ViewBinding.java:33) 
at butterknife.internal.DebouncingOnClickListener.onClick(DebouncingOnClickListener.java:22) 
at android.view.View.performClick(View.java:6205) 
at android.view.View$PerformClick.run(View.java:23653) 
at android.os.Handler.handleCallback(Handler.java:751) 
at android.os.Handler.dispatchMessage(Handler.java:95) 
at android.os.Looper.loop(Looper.java:154) 
at android.app.ActivityThread.main(ActivityThread.java:6682) 
at java.lang.reflect.Method.invoke(Native Method) 
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1520) 
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1410) 
```

miquelbeltran:
I submitted a change, I suspect the issue is that ViewModelDemo was not public but package-private. Surprisingly it worked for me all the time.

jshvarts:
Making `ViewModelDemo` public fixed the crash.

esafirm:
How one can share the `ViewModel` between parent `Controller` to its children?

miquelbeltran:
@esafirm I'd not share the same instance of a ViewModel across views (doesn't matter if they are Controllers, Activities or Fragments).

You could of course have different instances of the same ViewModel class on as much views as you need. (Example: you have a ViewModel for a user profile, you can use this ViewModel on different screens where you need to display the user profile, but each one will have its own instance).

The question is also frequently asked regarding sharing VM across activities: https://github.com/googlesamples/android-architecture-components/issues/29

As well, the best way to ensure that the same data is used on all instances of the same ViewModel is by having a data repository that can store any state or api data that you need to display, and the ViewModels use that data repository.

conductor-modules/arch-components-lifecycle/src/main/java/com/bluelinelabs/conductor/archlifecycle/ViewModelController.java
```diff
@@ -0,0 +1,17 @@
+package com.bluelinelabs.conductor.archlifecycle;
+
+import android.arch.lifecycle.ViewModelProvider;
+import android.arch.lifecycle.ViewModelStore;
+
+public abstract class ViewModelController extends LifecycleController {
+
+    private ViewModelStore viewModelStore = new ViewModelStore();
```
esafirm: I think `viewModelStore.clear()` should be called on `Controller` `onDestroy`

esafirm:
> The question is also frequently asked regarding sharing VM across activities: googlesamples/android-architecture-components#29

The issue is about sharing `ViewModel` between activities, which in comparison would be sharing a `ViewModel` between `Controller` on the same level. 

What i'm asking is sharing a `ViewModel` between a `Controller` and their children (added via `getChildRouter`) and i think it is a common/valid use case. It's the same as sharing between `Activity` and their `Fragment`s.

AFAIK it is even mentioned on `ViewModel` introduction. (Can communicate between Activity and their Fragments in the decouple way)

conductor-modules/arch-components-lifecycle/src/main/java/com/bluelinelabs/conductor/archlifecycle/ViewModelController.java
miquelbeltran: Agree, as the docs say: ```If an owner of this ViewModelStore is destroyed and is not going to be recreated, then it should call clear() on this ViewModelStore, so ViewModels would be notified that they are no longer used.``` https://developer.android.com/reference/android/arch/lifecycle/ViewModelStore.html#clear() I'd look to add that to the ViewModelController onDestroy method.

miquelbeltran:
OK, I never use child Controllers so I cannot go into details. As far as they can access the parent Controller instance, they should be able to use the same ViewModelProvider/ViewModelStore, then get the same ViewModel instance. Then I guess we need to make the viewModelProvider method public not protected. If you give it a try let me know!

miquelbeltran:
As I needed this for another project I made a small library: https://github.com/miquelbeltran/conductor-viewmodel which also provides the Lifecycle events to handle the LiveData observer subscription properly to unsubscribe then the view is destroyed.

tomblenz:
I'd make `viewModelProvider()` public.  With activities and fragments, a fragment can pass its activity as a provider to get the activity ViewModel for ease of sharing data between fragments and activities (google have an example of this use case [in their docs](https://developer.android.com/topic/libraries/architecture/viewmodel.html#sharing)).

eg. this is currently not possible because `viewModelProvider()` is protected.
```
val vm = (parentController as ViewModelController)
      .viewModelProvider()
      .get(HomeViewModel::class.java)
```

miquelbeltran:
I agree. I'll make the change!

tomblenz:
I've been having a play with this PR and I have to say, this appears to tick all of the boxes for AC+Conductor interop (with the above exception).  Good job mate.

I also was not able to reproduce the duplicated observer thing in either Fragment or Controller, it may have been fixed in AC 1.1.1 or something?

miquelbeltran:
Thanks! I am thinking about rebasing this and make it 1:1 to what I have in  https://github.com/miquelbeltran/conductor-viewmodel

I added my own LifecycleProvider. See in here: https://github.com/bluelinelabs/Conductor/pull/405/files#diff-b7e5adc69ad34fd475e7fad2af3275a7

Btw I am using the conductor-viewmodel library on a couple of production projects, it's working great :)

miquelbeltran:
I have rebased the work, and as well moved the code to the com.bluelinelabs.conductor.viewmodel package.

As said, I have added my own ViewModelLifecycleOwner because the one found in the library is not handling the required Lifecycle events correctly. It's important that `ON_DESTROY` happens in the `onDestroyView` and not in `onDestroy` because you want to stop updating the UI (i.e. stop observing LiveData) when the view is destroyed, not when the Controller is destroyed.

miquelbeltran:
Hello all, since there has been no movement for months and I have a working solution as an independent library: https://github.com/miquelbeltran/conductor-viewmodel I've decided to close this PR.

If you are interested in the included functionality please feel free to use my library in conjunction with Conductor!

johnjeremih:
@miquelbeltran I currently working on a 5 years old project that uses a conductor. Why should I use Conductor over fragment? Why not replace it with fragments? I have been asking myself this question, but I need an expert to tell me what they think about it.

PaulWoitaschek:
The question is: Why would you want to use fragments?
Imo the arch view models don't give any real benefit as controllers survive config changed anyways and thats the main selling point for the arch view models.

johnjeremih:
@PaulWoitaschek I'm trying to follow what Google recommends and try to have an MVVM architecture, but with conductors, you don't have that. I like how you can manage the data in the view model and retrieve it in any fragment. Now, I would like to know your perspective and how you see the new trend of MVVM architecture.

EricKuck:
It's important to note that MVVM existed far before Google released the component that they call a ViewModel. It's also very important to note that the thing they call a ViewModel is not an MVVM view model. It's a thing that survives configuration changes and nothing more. You can use it as an MVC controller, an MVP presenter, or a handy place to store random state fields. 

The ViewModel class is very badly named and has mislead countless developers into thinking it's needed for MVVM. It's not. Its sole purpose is to wallpaper over the fact that activities and fragments recreate on configuration change.

Given that Conductor Controllers do not have this issue, the class provides no value. You don't need a base class called ViewModel to have a ViewModel. You can roll your own with no real effort. What I've done in the past is create a base class with a CoroutineScope (rx works too, or livedata if you really had to) that cancels on Controller lifecycle events. That's really all you need.

PR #467, opened 2018-08-18
Implement DialogController

MkhytarMkhoian:
Implement DialogController like DialogFragment
I already have this code in production and its work like a charm

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/DialogController.java
@@ -1,0 +1,282 @@
+package com.bluelinelabs.conductor;
+
+import android.app.Dialog;
+import android.content.DialogInterface;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.Window;
+import android.view.WindowManager;
+
+import com.bluelinelabs.conductor.changehandler.SimpleSwapChangeHandler;
+
+public class DialogController extends RestoreViewOnCreateController
+  implements DialogInterface.OnShowListener,
+  DialogInterface.OnCancelListener, DialogInterface.OnDismissListener {
+
+  /**
+   * Style for {@link #setStyle(int, int)}: a basic,
+   * normal dialog.
+   */
+  public static final int STYLE_NORMAL = 0;
+
+  /**
+   * Style for {@link #setStyle(int, int)}: don't include
+   * a title area.
+   */
+  public static final int STYLE_NO_TITLE = 1;
+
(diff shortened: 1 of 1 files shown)


EricKuck:
This looks really well done, and I'll probably be using it myself in future projects. I'm not sure it belongs in the core project though, as it's kind of ancillary to the intent of the library. What do you think of publishing it as your own library? I'd be happy to link to it from the main readme of this project.

MkhytarMkhoian:
I think it's a good idea. I will write here when I'm done.

MkhytarMkhoian:
@EricKuck I'm done

https://github.com/MkhytarMkhoian/ConductorDialog

`implementation 'com.github.mkhytarmkhoian:conductor-dialog:1.0.0'`

mradzinski:
@MkhytarMkhoian Hey! Thanks for your PR. What's the immediate advantage of this over showing a standard alert dialog or using #248? Not saying it's bad or anything (actually looks really well done! Kudos to you!), just wondering the reasons behind the PR so I can choose to use your library. Thanks 👍

PaulWoitaschek:
It handles state restoring and integrates with the rest of conductor for targetController etc.

MkhytarMkhoian:
@mradzinski Hi! When you show a standard alert dialog, you don't have this dialog in you back stack. One of a case, when you turn on/off screen, your dialog is gone and not showing again. Also, you can't manage you dialog state.

Sometimes you need to implement a more complicated dialog with own logic, in that case, you need something like `DialogFragment` or `DialogController` in case of Conductor. `DialogController` it's like `Controller` that wrapping dialog, so you have all the features of `Controller`.

About #248. This implementation's not good enough for me, because of API (no support for the theme and styles e.t.), also not correct state handling and the bug with change handler.

PaulWoitaschek:
@mradzinski What is "the bug with the change handler"

Thanks for releasing :tada:!

MkhytarMkhoian:
@PaulWoitaschek @mradzinski  The bug in `.popChangeHandler(new SimpleSwapChangeHandler(false))`. Do not pass `false` into `SimpleSwapChangeHandler` because the view not removed.

return stub view in `onCreateView` method.
```
  return new View(getActivity());//stub view
```

PaulWoitaschek:
And what issues does that cause?

MkhytarMkhoian:
Stub view not remove from main ViewGroup container, so they left in the container until someone call removeAllViews() on a container.  In case of stub view, it's not dangerous, but not good anyway,  in some case, it may cause memory leaks, extra overdraw.

MkhytarMkhoian:
@EricKuck HI, what about to link it to the main readme of this project?

EricKuck:
Link added to the readme. Thanks!

ursusursus:
Hi, this hides the view underneath for me when I call pushController, am I doing something wrong?

PR #468, opened 2018-08-19
Adds popCurrentController(ControllerChangeHandler) and popController(Controller, ControllerChangeHandler)

beylerian:
### About

Hello, thank you for this nice framework.
I am currently using it for one of my apps and it is lovely 👏 

One case I ran into was when I needed to be able to pop the top controller but with a different `ChangeHandler` than what was setup previously, due to the dynamic design of the app I am working with. In this case, the particular controller is not a `root` controller but the top one. 

So I felt it could be nice to have a `popCurrentController(ControllerChangeHandler)` method as well since that would avoid the need for tagging and using `popToTag(String, ControllerChangeHandler)`  in my current case.
But there was only a  `popCurrentController()` method, so I suggest to add this as well as `popController(Controller, ControllerChangeHandler)` 

### Changes

So this PR changes two things:

1. Adds `popCurrentController(ControllerChangeHandler): boolean` and `popController(Controller, ControllerChangeHandler): boolean` in `Router.java`
This makes the pop method list look like this:

```java
popCurrentController(): boolean
>popCurrentController(ControllerChangeHandler): boolean
popController(Controller): boolean
>popController(Controller, ControllerChangeHandler): boolean
popToRoot(): boolean
popToRoot(ControllerChangeHandler): boolean
popToTag(String): boolean
popToTag(String, ControllerChangeHandler): boolean
```

2. Demos the functionality in the navigation demos,
by showing a `Go Back` and `Go Back and Override Handler` buttons.
`Go Back and Override Handler` demos changing the already set `HorizontalChangeHandler` to a `VerticalChangeHandler`

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Router.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Router.java
@@ -110,13 +110,25 @@ public abstract class Router {
     @SuppressWarnings("WeakerAccess")
     @UiThread
     public boolean popCurrentController() {
+        return popCurrentController(null);
+    }
+
+    /**
+     * Pops the top {@link Controller} from the backstack
+     *
+     * @param changeHandler The {@link ControllerChangeHandler} to handle this transaction
+     * @return Whether or not any {@link Controller}s were popped in order to get to the transaction
+     */
+    @SuppressWarnings("WeakerAccess")
+    @UiThread
+    public boolean popCurrentController(@Nullable ControllerChangeHandler changeHandler) {
         ThreadUtils.ensureMainThread();
 
         RouterTransaction transaction = backstack.peek();
         if (transaction == null) {
             throw new IllegalStateException("Trying to pop the current controller when there are none on the backstack.");
         }
-        return popController(transaction.controller);
+        return popController(transaction.controller, changeHandler);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


EricKuck:
Thanks so much for taking the time to do all of this so thoroughly! This functionality actually exists in the library already, in the form of `Controller.overridePushHandler()` and `Controller.overridePopHandler()`. Is there a use case for using these new methods that the existing functionality doesn't cover?

beylerian:
Hi @EricKuck thank you for the follow-up!

Indeed, the existing functionality does provide the ability to perform the same behavior. 
However, I felt that those additional methods would be nice to include, mainly for the sake of convenience. Although we can definitely use the `overridePopHandler`, the convenience method would also do two steps in one.

For instance, in a Controller, instead of doing:

```java
this.overridePopHandler(newChangeHandler);
getRouter().popCurrentController();
```

We can just do:

```java
getRouter().popCurrentController(newChangeHandler)
```

This becomes particularly nice if we are switching a pop behavior through a `Router`, for example following a result, and/or where a "one-liner" would look better. 

Otherwise, in my particular use case, I don't require a `popController(Controller, ControllerChangeHandler)`, but it seemed like a natural addition when going over the pop methods, since similar ones do exist for other cases, as: 

```java
popToRoot(): boolean
popToRoot(ControllerChangeHandler): boolean

popToTag(String): boolean
popToTag(String, ControllerChangeHandler): boolean

popCurrentController(): boolean
->popCurrentController(ControllerChangeHandler): boolean

popController(Controller): boolean
->popController(Controller, ControllerChangeHandler): boolean
```

PR #471, opened 2018-08-23
Add support for recursively setting ControllerChangeListeners on a ro…

abyrnes:
Adds support for recursively setting ControllerChangeListeners, meaning that any of the router's child controllers' child routers will automatically have the same change listener set.

An additional `addChangeListener` method was added to the `Router` class. The new method has a second boolean parameter which, when set to `true`, enables to behavior described above: `router.addChangeListener(changeListener, true)`.

The behavior for `router.addChangeListener(changeListener, false)` is equivalent to `router.addChangeListener(changeListener)`


conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
```diff
@@ -90,6 +91,7 @@
     private WeakReference<View> destroyedView;
     private boolean isPerformingExitTransition;
     private boolean isContextAvailable;
+    private final List<ControllerChangeListener> recursivelySetChangeListeners = new ArrayList<>();
```
MkhytarMkhoian: Please, rename `recursivelySetChangeListeners ` -> `changeListeners `

MkhytarMkhoian (commented):
Please, rename recursivelySetChangeListeners -> changeListeners

EricKuck:
Rather than having two lists (`changeListeners` and `recursivelySetChangeListeners`), how about we keep it at one list that has some type of wrapper object that contains both the changeListener and a boolean for whether or not it's recursive?

I'm also thinking it might be better if Controllers didn't have a list of listeners like that. Perhaps each router should keep a list that only contains its own changeListeners. When the listeners are ready to be used, the router could then grab all of its parent routers' changeListeners as well. So basically, Router.java:798 would go from:

`final ChangeTransaction transaction = new ChangeTransaction(to, from, isPush, container, changeHandler, new ArrayList<>(changeListeners));`

to 

`final ChangeTransaction transaction = new ChangeTransaction(to, from, isPush, container, changeHandler, new ArrayList<>(getAllChangeListeners()));`

Make sense?

abyrnes:
I've made the requested changes; creating a wrapper for the change listener/recursive pair.

EricKuck:
Looks great - last thing: instead of having a `setupFor()` method, how about putting that in `setHost()`? I think that'll be a cleaner solution going forward, as there are different scenarios where `setHost()` gets called (ex: state restoration after process death) and I don't want to miss calling `setupFor()` at some point in the future.

abyrnes:
Sounds good. So I get the `Router` from the `Controller` in `setHost`?

Also would you prefer to expose a method from `Router` returning a list of the wrapper object or a separate method returning only the recursive change listeners?

EricKuck:
Yep, I think using `controller.getRouter()` is a good way to do it.

No new methods should have to be exposed. The `getAllListeners()` method should be sufficient, and it should be able to stay private.

`ControllerHostedRouter`'s `getAllListeners` implementation could work something like this:

```
private List<ControllerChangeListener> getAllListeners() {
    ArrayList<ControllerChangeListener> listeners = new ArrayList<>(changeListeners);
    if (hostController.getRouter() != null) {
        listeners.addAll(hostController.getRouter().getAllListeners());
    }
    return listeners;
}
```

while `ActivityHostedRouter`'s implementation would skip the `hostController` portion of that code.

abyrnes:
The `getAllChangeListeners()` method in `Router` returns `List<ControllerChangeListener>`, not `List<ChangeListenerData>`. So there'd have to be an additional method to either expose a list of the wrapper class or a list of only the recursive listeners.

The former would look like this (in `ControllerHostedRouter#setHost`):
```java
Router router = controller.getRouter();
for (ChangeListenerData changeListener : router.getChangeListenerData()) {
    if (changeListener.recursive) {
        addChangeListener(changeListener.changeListener);
    }
}
```

or we could do
```java
Router router = controller.getRouter();
for (ChangeListenerData changeListener : router.getRecursiveChangeListeners()) {
    addChangeListener(changeListener);
}
```

EricKuck:
Ah sorry, I totally misread how your code worked! I don't believe there should actually be anything done in `setHost` after all. The motivation of the `getAllChangeListeners()` method was that it would get all of its own change listeners, as well as its ancestors' recursive change listeners. So maybe it should be `getAllChangeListeners(recursiveOnly: boolean)` or something like that? Then Router.java:798 would call `getAllChangeListeners(false)` and the `ControllerHostedRouter` implementation would call `getAllChangeListeners(true)` on its host.

abyrnes:
I think the code in `setHost` is still needed since that's ultimately where the change listeners are set on the child router, no?:
```java
// Router.java

List<ControllerChangeListener> getAllChangeListeners(boolean recursiveOnly) {
    final List<ControllerChangeListener> changeListeners = new ArrayList<>();
    for (ChangeListenerData changeListener : this.changeListeners) {
        if (!recursiveOnly || changeListener.recursive) {
            changeListeners.add(changeListener.changeListener);
        }
    }
    return changeListeners;
}
```

```java
// ControllerHostedRouter.java

final void setHost(@NonNull Controller controller, @NonNull ViewGroup container) {
    if (hostController != controller || this.container != container) {
        removeHost();

        if (container instanceof ControllerChangeListener) {
            addChangeListener((ControllerChangeListener)container);
        }

        hostController = controller;
        this.container = container;

        for (RouterTransaction transaction : backstack) {
            transaction.controller.setParentController(controller);
        }

        for (ControllerChangeListener changeListener : controller.getRouter().getAllChangeListeners(true)) {
            addChangeListener(changeListener);
        }

        watchContainerAttach();
    }
}
```

EricKuck:
With the `getAllChangeListeners` method being called when the change listeners are needed, they won't have to be added to each child router's internal list of change listeners. My thought behind having the get method is that each router can only contain the list of listeners that were explicitly added to it.

abyrnes:
Ohhh I think I finally see. `getAllChangeListeners` should return a list with all of its own routers as well as the recursively set ones in its root router?

EricKuck:
Not just root router, but all ancestors. The code I sent earlier would mostly do it. Should probably be something more like this though:


ControllerHostedRouter:
```
private List<ControllerChangeListener> getAllListeners(boolean recursiveOnly) {
    ArrayList<ControllerChangeListener> listeners = new ArrayList<>();
    for (ControllerChangeListenerData listener : controllerChangeListeners) {
        if (!recursiveOnly || listener.isRecursive) {
            listeners.add(listener.listener);
        }
    }

    if (hostController.getRouter() != null) {
        listeners.addAll(hostController.getRouter().getAllListeners());
    }
    return listeners;
}
```


ActivityHostedRouter:
```
private List<ControllerChangeListener> getAllListeners(boolean recursiveOnly) {
    ArrayList<ControllerChangeListener> listeners = new ArrayList<>();
    for (ControllerChangeListenerData listener : controllerChangeListeners) {
        if (!recursiveOnly || listener.isRecursive) {
            listeners.add(listener.listener);
        }
    }

    return listeners;
}
```

Perhaps a helper method would be better to reduce code duplication though. Your call.

abyrnes:
I actually no longer see a need for the `recursiveOnly` parameter as `getAllChangeListeners` is only called in one place. I think it would be sufficient for the base router class to return its internally maintained list of listeners via `getAllChangeListeners` and for `ControllerHostedRouter` to returns its own list in addition to the recursive listeners set on all ancestor routers and for `ActivityHostedRouter` to simply use the default implementation since its `getRootRouter` method just returns itself. So I think the following might be sufficient:

Router:
```java
List<ChangeListenerData> getAllChangeListenerData() {
    return new ArrayList<>(changeListeners);
}

List<ControllerChangeListener> getAllChangeListeners() {
    final List<ControllerChangeListener> changeListeners = new ArrayList<>();
    for (ChangeListenerData changeListener : this.changeListeners) {
        changeListeners.add(changeListener.changeListener);
    }
    return changeListeners;
}
```

ControllerHostedRouter:
```java
@Override
List<ControllerChangeListener> getAllChangeListeners() {
    List<ControllerChangeListener> changeListeners = getAllAncestorChangeListeners(getRootRouter(), new ArrayList<ControllerChangeListener>());
    changeListeners.addAll(super.getAllChangeListeners());
    return changeListeners;
}

private static List<ControllerChangeListener> getAllAncestorChangeListeners(Router router, List<ControllerChangeListener> changeListeners) {
    for (ChangeListenerData changeListenerData : router.getAllChangeListenerData()) {
        if (changeListenerData.recursive) {
            changeListeners.add(changeListenerData.changeListener);
        }
    }

    Router parentRouter = router.getRootRouter();
    if (parentRouter == router) {
        return changeListeners;
    }

    return getAllAncestorChangeListeners(parentRouter, changeListeners);
}
```

EricKuck:
The issue is that routers can be nested arbitrarily deeply. A ControllerHostedRouter can host another ControllerHostedRouter, which can host another ControllerHostedRouter, and so on. If it worked the way you described, routers nested a few levels deep would be getting access to listeners that they shouldn't have access to.

abyrnes:
> routers nested a few levels deep would be getting access to listeners that they shouldn't have access to

I think "shouldn't" might be subjective here. The idea for this feature stemmed from fragment manager lifecycle callbacks being able to be set recursively. That implementation does in fact set the callback all the way down the line. So an activity's fragments' child fragments child fragments will also get the callback set on them.

The problem I'm trying to solve involves performing some task for every controller in the app that implements a particular interface. Is there something I'm missing where this behavior can be attained with having a the listeners recursively set only one level deep?

The only other alternative I can think of off the top of my head is to provide a hook so a router an disallow itself from inheriting change listeners. But then again what if they want to inherit only _some_ of the listeners?

EricKuck:
The listeners I was referring to that child routers shouldn't have access to are the non-recursive listeners. Your implementation will give children access to non-recursive listeners that they shouldn't have. In the following example:

```
RouterA: ActivityHostedRouter
    - Listener1: non-recursive
    - Listener2: recursive

RouterB: ControllerHostedRouter (hosted in a controller from RouterA)
    - Listener3: non-recursive
    - Listener4: recursive

RouterC: ControllerHostedRouter (hosted in a controller from RouterB)
    - Listener5: non-recursive
    - Listener6: recursive
```

your implementation of `getAllChangeListeners()` will return Listener2, Listener3, Listener4, Listener 5, and Listener 6. It should not be returning Listener3.

There's one more issue with the code you sent in the previous message as well - the usage of `getRootRouter()`. That method returns the furthest ancestor, not the immediate ancestor, which is what you're looking for.

abyrnes:
Ok I think I've got it. Your past code snippets make a lost more sense now! Sorry I misunderstood :/
Gonna push up the changes

EricKuck:
Any update to this?

abyrnes:
Yea! Sorry. I made the changes you requested a while ago and did a squash. Looking at the code again I just saw a bug I had to fix. But hopefully it's good to go now.

The only other thing I can think of is reordering how the change listeners are retrieved so the ones that come from the parent router(s) are first in the list. What do you think?

PaulWoitaschek:
Has someone found a workaround for this?
I currently really need this functionallity and can't overwrite `getChildRouter` as it's final.

abyrnes:
I've just made a few small changes to this PR that include reordering the change listeners that are returned from calling `getAllChangeListeners()` and added tests.

Hopefully at @EricKuck will have some time to look over this and verify that everything is correct.

dimsuz:
Would be really nice to have this merged, we have to insert ugly workarounds to be able to use Conductor with [mosby-conductor](https://github.com/sockeqwe/mosby-conductor) and child controllers.

PR #474, opened 2018-08-27
Implement Bottom navigation tab with own back stack in each tab

MkhytarMkhoian:
@EricKuck I'll try to implement Bottom navigation tab with own back stack. It works, but it would be great if you review this code because I can miss some edge cases.

In #316 when you switch tab, the previous router destroyed all controllers in a stack, so if you try using MVP pattern it does not work because all presenters destroyed with controllers. In my implementation, it's alive and has the same lifecycle as a standard controller.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -238,6 +238,65 @@ public abstract class Controller {
         return childRouter;
     }
 
+    public interface TabControllerFactory {
+        RouterTransaction getRootControllerForTab();
+    }
+
+    private ControllerHostedRouter previousTabRouter = null;
+
+    @NonNull
+    public final Router switchChildRouterForTab(
+      @NonNull ViewGroup container,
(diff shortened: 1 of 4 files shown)


mateherber:
Hi! What is the status of this PR? Is there a possibility that it will be merged soon?

MkhytarMkhoian:
I think it's better to extract this feature to separated library, before that the code and idea need to be reviewed

mateherber:
Make sense. In the meantime I've tried it and wanted to use my own ChangeHandler for tab switchs (animation like in the guidelines) but the router detach executed earlier than changehandler perfomchange so the fromView it got was already null so views didn't get removed from container when switching tabs :(

sdelaysam:
@MkhytarMkhoian another option is to use `RouterPagerAdapter` with custom view pager implementation (which, unlike `androidx.viewpager.widget.ViewPager` will disallow scrolling and keep only one view created).
That way you will get required behavior for free without library modifications.

Please refer to https://github.com/bluelinelabs/Conductor/pull/565

PR #482, opened 2018-09-29
Added require methods to Controller

abyrnes:
Adds the following require methods to Controller: `requireView()`, `requireActivity()`, `requireResources()`, `requireApplicationContext()`, `requireParentController()`, and `requireTargetController()`.

Useful for avoiding nullability errors when accessing these things when they're guaranteed to be non-null.

Calling code in Kotlin will no longer require null checks or `!!`s.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -280,6 +280,15 @@ public abstract class Controller {
         return view;
     }
 
+    @NonNull
+    public final View requireView() {
+        final View view = getView();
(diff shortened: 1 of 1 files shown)


ghost:
@EricKuck this is a must-have for us, Kotlin users.

EricKuck:
IMO this will just lead to bad practices and assuming you won't get null when you might actually get it. I've been a Kotlin user since it was in alpha and I would definitely not consider this to be a must-have.

Why not just use 

val context = context ?: throw IllegalStateException()

or make a simple extension to handle this that utilizes Kotlin contracts? It just doesn't seem like something that belongs in the core project to me.

Issue #422, opened 2018-03-23
Issues with kotlin synthetic imports


imGurpreetSK:
I cannot use synthetic imports directly when referencing view widgets from controller.
Ideally, I should be able to access the views like `loginButton.text = "Hello"`, but am unable to do so. However, I can access the views via my layout id, ie like `loginView.loginButton.text = "Hello"`. On checking the loginButton, teh IDE takes me to `R.java` file and shows the compiled ID of the view. Is there something I'm missing?

inorichi:
Create an abstract `BaseController` that implements the interface `LayoutContainer`.

Add the required `containerView`:

```kt
override val containerView: View?
    get() = view
```

And clear the view cache when the controller's view is destroyed:

```kt
override fun onDestroyView(view: View) {
    super.onDestroyView(view)
    clearFindViewByIdCache()
}
```

danielocampo2:
Hi @inorichi, thanks for the hint but doesn't seem to be working for me. Still need to reference view.viewId. Am I missing something?
Also, I was wondering, what is that `clearFindViewByIdCache()` function? 

Thanks.

inorichi:
You may have to add this at the end of your `build.gradle`

```
androidExtensions {
    experimental = true
}
```

PaulWoitaschek:
The synthetic properties are no magic. Kotlins compiler does not know what a Controller is. 

Check https://kotlinlang.org/docs/tutorials/android-plugin.html

miquelbeltran:
@GurpreetSK95 you can access views through the `getView()` method, which becomes just `view` on Kotlin.

Then it looks like this:

`view?.my_text_view?.text = "text"`

You need to use `?.` since `getView` is Nullable.

If you don't like it, you can do the following too:

```
view?.run {
    my_text_view.text = "text"
}
```

OR use the proposed solution by @inorichi which also supports view caching.

You can watch my talk (self promo time) where I explain how to access views with Kotlin's KAE: https://youtu.be/rsmvQBGJYcM?t=15m43s (link opens the section where I explain it)

imGurpreetSK:
Hi! Thanks for all the feedback and help. Reverting to you on the same, the issue is I don't want to do `view.someView.something` everytime I'm accessing or setting something. (It kinda defeats the purpose?) Also, with @inorichi 's way, doesn't clearing view cache all an overhead of creating the same view everytime a controller is resumed?

inorichi:
[...]

PR #504, opened 2018-11-20
Adds override method for onViewCreated.

ghost:


EricKuck:
What's the use case for having a method like this?

ghost:
I can speak about a personal use case, say I want to tell my presenter to load some data when the view is created (which fetches from a local data source) and then display that data.

Currently to do this, i'd make the call to presenter before the onCreateView returns. Then once it comes back in my Controller have something like `view?.customView?.bind(data)` `view` being the View reference held by Controller, but if onCreateView has not returned that reference will still be null. 

Maybe is a bit of an edge case, but IMO would be nice to be able to kick off work with the assurance the view held by Controller has definitely been assigned.

inorichi:
It happened to me too with kotlin synthetic properties, but you can workaround this with a custom BaseController too:

```kt
abstract class BaseController(bundle: Bundle? = null) : RestoreViewOnCreateController(bundle) {

    init {
        addLifecycleListener(object : LifecycleListener() {
            override fun postCreateView(controller: Controller, view: View) {
                onViewCreated(view)
            }
        })
    }

    open fun onViewCreated(view: View) { }

}

**Comment by dimsuz** [comment]
I did this without adding lifecycle listener:

```kotlin
abstract class BaseController : Controller {

  constructor()
  constructor(args: Bundle) : super(args)

  final override fun onCreateView(inflater: LayoutInflater, container: ViewGroup): View {
    val rootView = inflater.inflate(getViewLayout(), container, false)
    initializeView(rootView)
    return rootView
  }

  abstract fun initializeView(rootView: View)
}
```

But with this implementation `initializeView` will have a view which is not yet attached to the view hierarchy, if this is important, than @inorichi solution is probably a better one.

inorichi:
That wouldn't work with the `LayoutContainer` interface from the synthetic properties because it uses `getView()`. But you could also store locally `rootView` and set it to null in `onDestroyView`.

[Here's](https://github.com/bluelinelabs/Conductor/issues/422#issuecomment-386910696) the full controller implementation I added in another issue.

ghost:
I have actually used both solutions in the past, prefer the one from @inorichi as you can be sure the view held by the Controller has been set. 

Although ideally I would prefer to remove the need for a workaround base class altogether.

dimsuz:
> That wouldn't work with the LayoutContainer interface 

Yep, I omitted those parts for brevity, but here's the full code which I use for working with synthetic properties:

```kotlin
abstract class BaseController : Controller, LayoutContainer {
  private var bindPropsRootView: View? = null

  constructor()
  constructor(args: Bundle) : super(args)

  final override fun onCreateView(inflater: LayoutInflater, container: ViewGroup): View {
    val rootView = inflater.inflate(getViewLayout(), container, false)
    bindPropsRootView = rootView
    // initialization can happen only after bindPropsRootView is assigned, this is required for view cache to work
    initializeView(rootView)
    return rootView
  }

  @LayoutRes
  abstract fun getViewLayout(): Int

  abstract fun initializeView(rootView: View)
  protected open fun destroyView(view: View) {}

  override val containerView: View? get() = bindPropsRootView

  final override fun onDestroyView(view: View) {
    clearFindViewByIdCache()
    bindPropsRootView = null
    super.onDestroyView(view)
  }
}
```

PR #507, opened 2018-12-16
Add RxLifecycle3 support.

sewar:


EricKuck:
Is RxLifecycle still in general use? I haven't really paid much attention to it since AutoDispose came out, but have things changed since this blog post? https://blog.danlew.net/2017/08/02/why-not-rxlifecycle

I'm mostly wondering if it's even worth bothering to continue supporting something that's basically deprecated.

sewar:
RxLifecycle is kind-of-deprecated but still maintained, version 3 is just version 2 for AndroidX. Some projects haven't migrated to other solutions yet.

Right now there's not much effort needed to continue supporting RxLifecycle, when big changes come to Conductor it can be dropped then.

PR #511, opened 2019-01-16
Ab/agp 3.0

abyrnes:
After having downloaded AS 3.3, doing a clean build would produce the following error:
```
Invoke-customs are only supported starting with Android O (--min-api 26)
```
Targeting Java 8 fixed the problem

Also upgraded AGP to 3.3.0, which also required butterknife to be updated due to the following error:
```
The given artifact contains a string literal with a package reference 'android.support.v4.content' that cannot be safely rewritten. for androidx
```
Updating ButterKnife fixes this problem since 10.0.0 targets only androidx

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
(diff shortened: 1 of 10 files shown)


abyrnes:
Incorrect branch name + commit message :/

PR #513, opened 2019-01-16
Add ControllerFactory.

### LINKED ISSUE #510: Controller Constructor Injection
Currently, when injecting dependencies into a `Controller`, we can't perform constructor injection because Conductor needs to be able to recreate the Controllers after a configuration change. This sucks because field injection is confusing and error-prone.

Can we provide an abstraction like a `ControllerFactory` which would behave similarly to `androidx.work.WorkerFactory` in that consumers could override it with their own `ControllerFactory`?

AngusMorton:
ControllerFactory allows Controllers to be provided dependencies through their constructors after a configuration change. This means field injection is no longer a requirement when injecting dependencies into a Controller.

**Usage**
```
Conductor.setControllerFactory(myControllerFactory)
```

Currently, you have a single `ControllerFactory` for your entire app. Having a ControllerFactory per `Activity/ViewGroup` would be better but I don't know enough about Conductors internals to make that change.

Closes #510.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -92,14 +92,13 @@ public abstract class Controller {
     private boolean isPerformingExitTransition;
     private boolean isContextAvailable;
 
+    static volatile ControllerFactory FACTORY = DefaultControllerFactory.INSTANCE;
+
     @NonNull
     static Controller newInstance(@NonNull Bundle bundle) {
         final String className = bundle.getString(KEY_CLASS_NAME);
         //noinspection ConstantConditions
         Class cls = ClassUtils.classForName(className, false);
-        Constructor[] constructors = cls.getConstructors();
-        Constructor bundleConstructor = getBundleConstructor(constructors);
-
         Bundle args = bundle.getBundle(KEY_ARGS);
         if (args != null) {
             args.setClassLoader(cls.getClassLoader());
@@ -107,15 +106,20 @@ public abstract class Controller {
 
         Controller controller;
         try {
-            if (bundleConstructor != null) {
-                controller = (Controller)bundleConstructor.newInstance(args);
-            } else {
-                //noinspection ConstantConditions
-                controller = (Controller)getDefaultConstructor(constructors).newInstance();
+            controller = FACTORY.create(className, args);
+            if (controller == null) {
+                Constructor[] constructors = cls.getConstructors();
+                Constructor bundleConstructor = getBundleConstructor(constructors);
+                if (bundleConstructor != null) {
+                    controller = (Controller) bundleConstructor.newInstance(args);
+                } else {
+                    //noinspection ConstantConditions
+                    controller = (Controller) getDefaultConstructor(constructors).newInstance();
 
-                // Restore the args that existed before the last process death
-                if (args != null) {
-                    controller.args.putAll(args);
+                    // Restore the args that existed before the last process death
+                    if (args != null) {
+                        controller.args.putAll(args);
+                    }
                 }
             }
         } catch (Exception e) {
@@ -1321,6 +1325,10 @@ public abstract class Controller {
     }
 
     private void ensureRequiredConstructor() {
+        if (FACTORY != DefaultControllerFactory.INSTANCE) {
+            // We assume you know what you're doing if you've installed a ControllerFactory
+            return;
+        }
         Constructor[] constructors = getClass().getConstructors();
         if (getBundleConstructor(constructors) == null && getDefaultConstructor(constructors) == null) {
             throw new RuntimeException(getClass() + " does not have a constructor that takes a Bundle argument or a default constructor. Controllers must have one of these in order to restore their states.");
--- a/conductor/src/main/java/com/bluelinelabs/conductor/ControllerFactory.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/ControllerFactory.java
@@ -1,0 +1,10 @@
+package com.bluelinelabs.conductor;
+
+import android.os.Bundle;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+public interface ControllerFactory {
+    @Nullable
+    Controller create(@NonNull String controllerName, @Nullable Bundle args);
+}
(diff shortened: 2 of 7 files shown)


conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
```diff
@@ -92,6 +92,8 @@
     private boolean isPerformingExitTransition;
     private boolean isContextAvailable;
 
+    static volatile ControllerFactory FACTORY = new ControllerFactory.DefaultControllerFactory();
```
AngusMorton: I'm not sure if this is the best place for this to live. Maybe there is a better place for it? Does it need to be static? Would make sense for a `Router` to have a `ControllerFactory`?

conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
```diff
@@ -1320,7 +1322,8 @@ final void setParentController(@Nullable Controller controller) {
 
     private void ensureRequiredConstructor() {
         Constructor[] constructors = getClass().getConstructors();
-        if (getBundleConstructor(constructors) == null && getDefaultConstructor(constructors) == null) {
+        if (getBundleConstructor(constructors) == null && getDefaultConstructor(constructors) == null
+                && FACTORY instanceof ControllerFactory.DefaultControllerFactory) {
```
AngusMorton: If the user has installed a `ControllerFactory` we assume they know what they're doing. Maybe we should do some deeper/more expensive checking here? e.g. check that the installed `ControllerFactory` can create the `Controller`?

PaulWoitaschek:
I would make the Factory an interface and make the default factory final and hide the implementation.
That doesn't force the user to extend a class (composition over inheritance).

Then you could also just do a identity check in `ensureRequiredConstructor()` and perform an early return without the need for reflection:
```java
if(FACTORY != DefaultControllerFactory.instance() {
  return
}
```

Another approach (which I prefere) would be to pass the factory to conductor upon initialzation:
`Conductor.attachRouter(this, controllerContainer, savedInstanceState, factory)`

Then in `attachRouter` you can just set the factory and don't expose the other implementations.
I don't have currently time to check if this could work.

But I really like the idea to get rid of reflection and allow constructor injection.
However this whole approach is quite limited as dagger has no support for assistend injection. That effectively means that once need a bundle in the constructor you either need to do field injection or workaround it.

AngusMorton:
Thanks for the feedback! I've made the interface change.

If `attachRouter` accepts a factory, then it follows that you should be able to set a different `ControllerFactory` per `Activity`/`ViewGroup` pair. 
I think having a ControllerFactory per `Activity`/`ViewGroup` makes sense, but I'm not sure if it'll complicate things too much though.

Dagger doesn't need first-party support for assisted injection, it can be provided through something like [AssistedInject](https://github.com/square/assistedinject) or [AutoFactory](https://github.com/google/auto/tree/master/factory).

conductor/src/main/java/com/bluelinelabs/conductor/ControllerFactory.java
```diff
@@ -0,0 +1,10 @@
+package com.bluelinelabs.conductor;
+
+import android.os.Bundle;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+public interface ControllerFactory {
+    @Nullable
+    Controller create(@NonNull String controllerName, @Nullable Bundle args);
```
AngusMorton, line 9: Is `controllerClass` or `controllerName` more useful here? e.g. `Controller create(Class controllerClass, Bundle args);`

conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
```diff
@@ -92,30 +92,34 @@
     private boolean isPerformingExitTransition;
     private boolean isContextAvailable;
 
+    static volatile ControllerFactory FACTORY = DefaultControllerFactory.INSTANCE;
```
MkhytarMkhoian, line 95: Controller shouldn't know about any factory

conductor/src/main/java/com/bluelinelabs/conductor/DefaultControllerFactory.java
```diff
@@ -0,0 +1,16 @@
+package com.bluelinelabs.conductor;
+
+import android.os.Bundle;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+final class DefaultControllerFactory implements ControllerFactory {
```
MkhytarMkhoian, line 7: I think DefaultControllerFactory should implement default logic for creating Controller, but not just be a stub in logic

AngusMorton: We could only do that without extra reflection if the first argument for `create` was a `Class`.

conductor/src/main/java/com/bluelinelabs/conductor/ControllerFactory.java
AngusMorton: We can't make it Class because of Split APKs.

AngusMorton:
@PaulWoitaschek I had a look into adding `ControllerFactory` as an argument to `attachRouter`, however, we need some way to retrieve the correct `ControllerFactory` for the `Activity` when we are restoring the `Backstack`. I'm not familiar enough with the internals of Conductor to know what the right approach for this would be so I'm keen to hear any thoughts you have.

Assuming we store the ControllerFactory in the LifecycleHandler, we would need some way for a `Router` to get it's `ControllerFactory`, which it would then pass that down through to `Backstack.restoreInstanceState` -> `new RouterTransaction`.

AngusMorton:
@EricKuck I'm keen to hear your thoughts on this and move this forward

Miha-x64:
One more thing: if Factory instantiates Controllers, we need a way to create transaction without instantiating controllers ourselves. `RouterTransaction.with(controllerType: Class<out Controller>)`

AngusMorton:
> One more thing: if Factory instantiates Controllers, we need a way to create transaction without instantiating controllers ourselves. `RouterTransaction.with(controllerType: Class<out Controller>)`

`ControllerFactory` would be in your dependency graph so you can inject it and call it yourself!

PR #550, opened 2019-07-11
Keep ControllerChangeHandler default constructor in Proguard rules

DSteve595:
They can currently be stripped out, causing `ensureDefaultConstructor()` to fail.

### DIFF
--- a/conductor/proguard-rules.txt
+++ b/conductor/proguard-rules.txt
@@ -3,3 +3,6 @@
    public <init>();
    public <init>(android.os.Bundle);
 }
+-keepclassmembers public class * extends com.bluelinelabs.conductor.ControllerChangeHandler {
+   public <init>();
+}


conductor/proguard-rules.txt
```diff
@@ -3,3 +3,6 @@
    public <init>();
    public <init>(android.os.Bundle);
 }
+-keepclassmembers public class * extends com.bluelinelabs.conductor.ControllerChangeHandler {
+   public <init>();
+}
```
dimsuz, line 8: Oh, yes, we have to add this in each project which uses Conductor.
Also, this seems to be strictly required when using R8-shrinker (the new default), older projects somehow work without this on Proguard, still wouldn't hurt to have this always.

The only thing our rule differs in is we have `public <init>(...);`
Maybe this'll be more future proof if argument set ever changes, though I'm not sure it will happen.

PR #565, opened 2019-09-10
#260: bottom navigation demo

sdelaysam:
Another attempt to make a demo for #260
Also check https://github.com/bluelinelabs/Conductor/pull/316 and https://github.com/bluelinelabs/Conductor/pull/474


fonix232:
May I recommend a more elegant, generic solution that is completely self-contained? It's in Kotlin though:

https://gist.github.com/fonix232/14294caea86c4161478f5263a41fc50b

I've used the generic ViewPager, and have not prepared to swipes, but it should be relatively straightforward to use your `NonScrollableViewPager` implementation.

sdelaysam:
@fonix232 any good reason not to use the `RouterPagerAdapter` provided by Conductor?

fonix232:
@sdelaysam Using Kotlin I've simplified the logic of the RouterPagerAdapter. It's basically a verbatim copy, with some Kotlin-specific sugar added on top. No other reason whatsoever, in fact if you want me to base one off of RouterPagerAdapter, I can probably do it in a few minutes.

sdelaysam:
@fonix232 The whole project including demos using Java so I'm not sure about pushing Kotlin code. And it seems our PRs won't be merged anytime soon here :sweat_smile:
Anyway, thanks for sharing Kotlin version!

fonix232:
Honestly, apart from the handful Java purist crybabies, I don't see any valid reason why Conductor couldn't and shouldn't transition to Kotlin soon.  The codebase would be cut in about half, would be nicer to use and read, and so on.

vincent-paing:
I've added a solution that doesn't require `ViewPager` and it works with just a `ViewGroup`. I'm currently using it in my production app but I can't guarantee 100% works without any test coverage

https://gist.github.com/vincent-paing/4a49bd85e21689d069deb36374112be0

EDIT: There's some issue in restoring state, and do not use this code

PR #569, opened 2019-11-03
Fix #563 bug: Controller in the middle of a backstack isn't destroyed after calling popController

### LINKED ISSUE #563: Controller in the middle of a backstack isn't destroyed after calling popController
When I call popController for a controller in the middle of a backstack the Controller.performDestroy isn't called but the view of the controller is removed and the controller is removed from the backstask.

As popController has the flow for false poppingTopController I suppose everything should have worked correctly.

I've read the code and there are my thoughts about that problem:
In poppingTopController we call performControllerChange with null nextTransaction which I think is right. Then this popController leads to AnimatorChangeHandler.performAnimation call which seems strange as I don't expect any animation runned in this case. Or AnimatorChangeHandler.needsImmediateCompletion should be set to true from the router. In performControllerChange I see we set forceDetachDestroy to true only in a special case but my case might fit here too as it seems that would fix the issue. Then the view is removed when the animation finishes and as isBeingDestroyed is false the controller isn't destroyed.

There is the fork with a bit modified "Navigation Demos": https://github.com/soniccat/Conductor. I've added "Pop 3" button where I set a lifecycle listener to catch preDestroy. Then I trigger popping the 3rd controller. So, to reproduce the issue you need open, for example, 5 controllers and press this "Pop 3" button and see that preDestroy is not called.

I'd try to create a pull request if what I wrote above make sense :)

soniccat:
When I call popController for a controller in the middle of a backstack the Controller.performDestroy isn't called but the view of the controller is removed and the controller is removed from the backstask.

As popController has the flow for false poppingTopController I suppose everything should have worked correctly.

The original issue is here: https://github.com/bluelinelabs/Conductor/issues/563

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Router.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Router.java
@@ -146,9 +146,7 @@ public abstract class Router {
             while (iterator.hasNext()) {
                 RouterTransaction transaction = iterator.next();
                 if (transaction.controller == controller) {
-                    if (controller.isAttached()) {
-                        trackDestroyingController(transaction);
-                    }
+                    trackDestroyingController(transaction);
                     iterator.remove();
                     removedTransaction = transaction;
                 } else if (removedTransaction != null) {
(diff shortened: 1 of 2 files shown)


dimsuz:
Shouldn't there be a new unit-test written along with the fix? If I'm not mistaken this project has tests for all corner cases because there are a lot of them.

soniccat:
@dimsuz I'll make it after getting any response from any contributors. Probably the fix is wrong... or it's by design that controllers aren't destroyed in this scenario...

EricKuck:
A cursory glance at the PR looks great. I'll give it a better look through later day. Tests would be a great addition though. Even if your fear that your fix is "wrong" was correct, having the tests would help validate whatever the correct fix ended up being.

soniccat:
@EricKuck I see, I've added a new one

Issue #287, opened 2017-05-08
onAttach with isDestroyed=true


PaulWoitaschek:
I have a use case where when the user clicks a tab in a view pager, it pushes another controller instead an resets the position of the view pager.

Now with 2.1.3 I receive crashes because onAttach of a controller is being called after it is destroyed (its onDestroyView) was called and I nulled a variable in onDestroyView I now access in onAttach.

```kotlin
class MainActivity : AppCompatActivity() {

  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    val container = findViewById(android.R.id.content) as ViewGroup
    val router = Conductor.attachRouter(this, container, savedInstanceState)
    if (!router.hasRootController()) {
      router.setRoot(RouterTransaction.with(RootController()))
    }
  }

  class RootController : Controller() {
    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup): View {
      val pager = ViewPager(activity!!)
      pager.adapter = Adapter(this)


      Handler().postDelayed({
        pager.setCurrentItem(2, true)
        router.pushController(RouterTransaction.with(LoggingController()).pushChangeHandler(FadeChangeHandler()))
        pager.setCurrentItem(0, true)
      }, 2000)

      return pager
    }
  }

  class Adapter(host: Controller) : RouterPagerAdapter(host) {
    override fun configureRouter(router: Router, position: Int) {
      if (!router.hasRootController()) {
        router.setRoot(RouterTransaction.with(LoggingController()))
      }
    }

    override fun getCount() = 3
  }

  class LoggingController : Controller() {
    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup) = View(activity!!).apply {
      val rnd = Random()
      val color = Color.argb(255, rnd.nextInt(256), rnd.nextInt(256), rnd.nextInt(256))
      setBackgroundColor(color)
    }

    override fun onAttach(view: View) {
      Log.d(toString(), "onAttach with isDestroyed=$isDestroyed")
    }

    override fun onDestroyView(view: View) {
      Log.d(toString(), "onDestroyView")
    }
  }
}
```

This logs:
```
D/de.paul_woitaschek.conductoronattachbug.MainActivity$LoggingController@11c05d2: onAttach with isDestroyed=false
D/de.paul_woitaschek.conductoronattachbug.MainActivity$LoggingController@6becda0: onAttach with isDestroyed=false
[...]

PR #570, opened 2019-11-03
Fix Conductor pending changes lock, related to #287 discussion

soniccat:
According my comment in https://github.com/bluelinelabs/Conductor/issues/287 here I've fixed locking all changes to a router because of having constantly non empty pendingControllerChanges after rotation on launching.

That can be reproduced after adding this lines in the demo activity and launching app in a rotated simulator: https://github.com/soniccat/Conductor/commit/a86d0c6428e6bed412dc46b2e7d720d7d7d204e9


soniccat:
Added the test for this case and another one for Router.pendingControllerChanges

Issue #518, opened 2019-02-27
ChildController's in viewpager, onDetach, onDestroy called 


tonyjs:
first of all here is router pager adapter.
```
class PageAdatper : RouterPagerAdapter {
    override configureRouter(router: Router, position: Int) {
        if (!router.hasRootController()) {
            when (position) {
                0 -> router.setRoot(RouterTransaction.with(AController())
                1 -> router.setRoot(RouterTransaction.with(BController()))
                2 -> router.setRoot(RouterTransaction.with(CController()))
        }
```

now open `TestController` inside on `AController`
```
class AController : Controller {
    ...
    override onViewBound(view: View) {
        view.btn_test.onClick {
            getChildRouter(view.test_view_group).apply { setPopsLastView(true) }
                .pushController(RouterTransaction.with(TestController()))
        }
    }
    ...
}
```

then I pressed the back button on `TestController`,
i can see the log onDetach and onDestroy called in `BController` and `CController`

is this a bug? or intended?

if intended, which is recommended way to open inside controller on parent controller which insside viewpager?

dimsuz:
I believe this happens due to the same reasons described in #314 

Namely: Conductor has a bug where it detaches child controllers even if parent controller has RETAIN_DETACH set - and never attaches them back afterwards.

@PaulWoitaschek  @EricKuck should this be closed as a duplicate?

tonyjs:
yeah, similar. 
I've found the solution(confused, It is right way), 
do not push controller with child router inside viewpager(or already exist on child router), use parent router(maybe root router). and handle the backstack.
E.g
```
override fun handleBack() {
    return pageAdapter.getRouter(pager.currentItem).handleBack()
}
```

if making child router, must to manage the backstack. maybe, attach backstack into parent router's backstack.

EricKuck:
Fixed via #571

Issue #314, opened 2017-06-12
Lifecycle issue with RETAIN_DETACH-enabled ViewPager host controller and child controllers


dimsuz:
When host controller of `RouterPagerAdapter` has `RETAIN_DETACH` mode set and another controller is being pushed on top of the host controller, then `onDetach` gets called on individual child page controllers inside the adapter, but after popping back to host controller, `onAttach` never gets called on those page controllers. This breaks `onAttach`/`onDetach` symmetry and I suspect is wrong. At least I have problems with this while using a `mosby-conductor` library (it never reattaches view back to a presenter)

I prepared a [repository clone](https://github.com/dimsuz/Conductor/tree/router_pager_no_attach_bug) with a slightly modified `demo` app which can be used to reproduce the problem.
The diff of my changes can be seen in [this commit](https://github.com/dimsuz/Conductor/commit/8cf2d884e83b9061415a85ac6fb311293f05cc17).

Here are the steps to reproduce with this demo app:

1. Launch the demo
2. Go to `ViewPager` menu item
3. Click on the page title (it'll launch a `TextController` on top of the parent (host) controller)
4. Observe the logs:
```
Child Controller Child #0 (Swipe to see more): onDetach()
Child Controller Child #0 (Swipe to see more): onDestroyView()
```
5. Press "BACK" button to pop `TextController` and return to `PagerController`
6. Observe that logs have no calls to `onAttach` or `onCreateView` for `ChildController`...

If you remove `RETAIN_DETACH` mode then everything starts to work and `onAttach` will appear in logs on step 6 above.
Note that `RETAIN_DETACH` is set only on host controller (`PagerController`), child controllers have default behavior, but still their lifecycle ends up being broken.

dimsuz:
I seem to have found a reason why this happens.

When `PagerController` is set to use the `RETAIN_DETACH` mode, its view doesn't get destroyed, but its **child** controllers do get their views destroyed. Child controllers in this case happen to be controllers of pages which reside in `ControllerPagerAdapter`.

So when pushing a new Controller on top of `PagerController` with `RETAIN_DETACH` we get this picture:

`PagerController::onDestroyView` - not called
`ChildController::onDestroyView` (pages) - **called**

so this is kinda asymmetric. parent lives while its children die :)
And also later children will never get attached back. I think this is because parent view wasn't destroyed, so it doesn't care to reattach children, but i'm not sure.

[...]

PR #571, opened 2019-11-08
Fixed having a not attached child controller after popping back to the parent

soniccat:
Came across a bug when we have a parent controller with RETAIN_DETACH and its child controller with RELEASE_DETACH. If we push another RELEASE_DETACH controller to the parent router and then pop back, the child controller won't be attached and it will have a null view property but the view itself will be still in its parent and therefore visible... Here I've fixed the bug and added a correspond test.

The current workaround for this bug is setting RETAIN_DETACH for the child controller.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -922,6 +922,8 @@ public abstract class Controller {
                     childTransaction.controller.attach(childTransaction.controller.view);
                 }
             }
+
+            childRouter.rebindIfNeeded();
         }
     }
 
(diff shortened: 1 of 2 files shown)


dimsuz:
There are a few issues already reported about similar bugs. I found the one I reported: #314.
It doesn't require pushing RELEASE_DETACH-ed controller to reproduce, but still sounds really similar, perhaps you've fixed it along the way too.

dimsuz:
One more issue sounding alike this one: #518

dimsuz:
They both are mentioning ViewPager, but since reporting I have observed the same behaviour with plain Parent-Child controllers, so it seems it's this relationship that matters, unlikely that a ViewPager plays some major role here.

EricKuck:
@soniccat Thanks for all of these great bug fixes! Sorry it took so long to get to them, but I'm really happy they're there.

PR #584, opened 2020-03-26
Add androidx-transition module, update demos

DSteve595:
Added a new module, `androidx-transition`. Code is identical to `TransitionChangeHandler`, but using [AndroidX Transition](https://developer.android.com/jetpack/androidx/releases/transition) rather than framework `Transition`s.

Also updated the demos to use these in favor of `TransitionChangeHandlerCompat`s.


conductor/src/main/java/com/bluelinelabs/conductor/changehandler/SharedElementTransitionChangeHandler.java
```diff
@@ -17,17 +17,20 @@
...
+import com.bluelinelabs.conductor.internal.LegacyTransitionUtils;
 
 import java.util.ArrayList;
 import java.util.List;
 
 /**
  * A TransitionChangeHandler that facilitates using different Transitions for the entering view, the exiting view,
  * and shared elements between the two.
+ *
+ * @deprecated See {@link TransitionChangeHandler}
```
EricKuck, line 29: I don't think this link is going to work correctly. Maybe update it with the package name?

DSteve595: Works since it's in the same package.

(To clarify, this is referencing the existing `TransitionChangeHandler`, I just didn't want to duplicate the deprecation message.)

EricKuck: Oh, whoops. I misread this and thought you were linking to the androidx version.

PR #585, opened 2020-03-28
Introduce Kotlin (and Java 8)

### LINKED ISSUE #578: Java8 + Default Methods
I'd like to bump this library to java 8 and make use of default implementations.
For example a candidate where I most of the time only need one method to be overwritten is `ControllerChangeListener`.

@EricKuck Would you accept such a PR?

Or should we directly migrate to kotlin?

PaulWoitaschek:
Fixes  #578


conductor/src/main/java/com/bluelinelabs/conductor/ChangeHandlerFrameLayout.kt
```diff
@@ -0,0 +1,39 @@
...
+ * [ControllerChangeHandler]s are performing changes. It is not required to use this
+ * ViewGroup, but it can be helpful.
+ */
+class ChangeHandlerFrameLayout : FrameLayout, ControllerChangeListener {
+
+    private var inProgressTransactionCount = 0
+
+    constructor(context: Context) : super(context)
+    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs)
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
```
dimsuz: Perhaps it's better to use 

```kotlin
class ChangeHandlerFrameLayout @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null)
```
instead of manually specifying each constructor?

conductor/src/main/java/com/bluelinelabs/conductor/Conductor.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.app.Activity
+import android.os.Bundle
+import android.view.ViewGroup
+import androidx.annotation.UiThread
+import com.bluelinelabs.conductor.internal.LifecycleHandler
+import com.bluelinelabs.conductor.internal.ensureMainThread
+
+/**
+ * Point of initial interaction with Conductor. Used to attach a [Router] to your Activity.
+ */
```
dimsuz: This comment is not attached to any syntactic entity...

conductor/src/main/java/com/bluelinelabs/conductor/Conductor.kt
```diff
@@ -0,0 +1,35 @@
...
+ * for restoring the Router's state if possible.
+ * @return A fully configured [Router] instance for use with this Activity/ViewGroup pair.
+ */
+@UiThread
+fun Activity.attachRouter(container: ViewGroup, savedInstanceState: Bundle?): Router {
+    ensureMainThread()
+    val lifecycleHandler = LifecycleHandler.install(this)
+    val router = lifecycleHandler.getRouter(container, savedInstanceState)
+    router.rebindIfNeeded()
+    return router
```
dimsuz: this is probably a personal preference but slightly shorter way to do this would be

```kotlin
return lifecycleHandler.getRouter(container, savedInstanceState).apply {
  rebindIfNeeded();
}
```

conductor/src/main/java/com/bluelinelabs/conductor/RouterTransaction.kt
```diff
@@ -0,0 +1,112 @@
...
+ */
+class RouterTransaction
+private constructor(
+        @get:JvmName("controller")
+        val controller: Controller,
+        private var tag: String? = null,
+        private var pushControllerChangeHandler: ControllerChangeHandler? = null,
+        private var popControllerChangeHandler: ControllerChangeHandler? = null,
+        private var attachedToRouter: Boolean = false,
+        @RestrictTo(LIBRARY)
```
dimsuz: can this be replaced by "internal" visibility? (not sure)

conductor/src/main/java/com/bluelinelabs/conductor/internal/RouterRequiringFunc.kt
```diff
@@ -0,0 +1,9 @@
+package com.bluelinelabs.conductor.internal
+
+import androidx.annotation.RestrictTo
+import androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP
+
+@RestrictTo(LIBRARY_GROUP)
+interface RouterRequiringFunc {
+    fun execute()
+}
```
dimsuz, line 9: Can this be replaced by a typealias?

```kotlin
typealias RouterRequiringFunc = () -> Unit
```

conductor/src/main/java/com/bluelinelabs/conductor/ChangeHandlerFrameLayout.kt
PaulWoitaschek: I generally strongly dislike JvmOverloads for View constructors.
Different constructors might have different behavior (and if they dont have right now that might change in a future update).

conductor/src/main/java/com/bluelinelabs/conductor/Conductor.kt
PaulWoitaschek: Is it possible to add a kind of file description in kdoc?

conductor/src/main/java/com/bluelinelabs/conductor/Conductor.kt
PaulWoitaschek: Totally!

conductor/src/main/java/com/bluelinelabs/conductor/internal/RouterRequiringFunc.kt
PaulWoitaschek: In the long term of cause but as it won't work with java interop, we an only replace this after all usages migrated to kotlin.

conductor/src/main/java/com/bluelinelabs/conductor/RouterTransaction.kt
PaulWoitaschek: Unfortunatelly not. Internal becomes public on java.

conductor/src/main/java/com/bluelinelabs/conductor/Conductor.kt
PaulWoitaschek: I don't think so. I deleted it for now.

conductor/src/main/java/com/bluelinelabs/conductor/ChangeHandlerFrameLayout.kt
dimsuz: I see, ok then.

conductor/src/main/java/com/bluelinelabs/conductor/RouterTransaction.kt
```diff
@@ -0,0 +1,123 @@
...
+private const val KEY_INDEX = "RouterTransaction.transactionIndex"
+private const val KEY_ATTACHED_TO_ROUTER = "RouterTransaction.attachedToRouter"
+
+/**
+ * Metadata used for adding [Controller]s to a [Router].
+ */
+class RouterTransaction
+private constructor(
+  @get:JvmName("controller")
+  val controller: Controller,
```
EricKuck, line 22: Perhaps we should annotate this with `@JvmField` so Java users don't have to convert all their `transaction.controller` usages to `transaction.controller()`?

conductor/src/main/java/com/bluelinelabs/conductor/ControllerChangeType.kt
```diff
@@ -0,0 +1,18 @@
+package com.bluelinelabs.conductor
+
+/**
+ * All possible types of [Controller] changes to be used in [ControllerChangeHandler]s
+ */
+enum class ControllerChangeType(val isPush: Boolean, val isEnter: Boolean) {
```
EricKuck: `@JvmField` would be nice on `isPush` and `isEnter` as well.

conductor/src/main/java/com/bluelinelabs/conductor/Router.java
```diff
@@ -325,7 +324,7 @@ public boolean popToTag(@NonNull String tag, @Nullable ControllerChangeHandler c
     }
 
     /**
-     * Sets the root Controller. If any {@link Controller}s are currently in the backstack, they will be removed.
+     * Sets the root controller(). If any {@link Controller}s are currently in the backstack, they will be removed.
```
EricKuck: Looks like a copy pasta error. Should still be `Controller`.

conductor/src/main/java/com/bluelinelabs/conductor/Router.java
```diff
@@ -796,7 +790,7 @@ private void performControllerChange(@Nullable RouterTransaction to, @Nullable R
             // Activity or controller should be handling this by finishing or at least hiding this view.
             changeHandler = new NoOpControllerChangeHandler();
             forceDetachDestroy = true;
-        } else if (!isPush && fromController != null && !fromController.isAttached()) {
+        } else if (!isPush && fromController != null && !from.controller().isAttached()) {
```
EricKuck: Why not use the local `fromController` variable anymore?

conductor/src/main/java/com/bluelinelabs/conductor/Router.java
```diff
@@ -805,10 +799,10 @@ private void performControllerChange(@Nullable RouterTransaction to, @Nullable R
         performControllerChange(toController, fromController, isPush, changeHandler);
 
         if (forceDetachDestroy && fromController != null) {
-            if (fromController.getView() != null) {
-                fromController.detach(fromController.getView(), true, false);
+            if (from.controller().getView() != null) {
+                from.controller().detach(from.controller().getView(), true, false);
             } else {
-                fromController.destroy();
+                from.controller().destroy();
```
EricKuck: Why not use the local `fromController` variable anymore?

EricKuck:
This looks great to me. I left a few small comments and questions, but it should be ready to merge soon. Thanks for all the work!

conductor/src/main/java/com/bluelinelabs/conductor/RouterTransaction.kt
PaulWoitaschek: No, the field `controller` was package private. There was a method called `.controller()` so the JvmName keeps these usages the same. It's only the internal usages which are changing.

PR #587, opened 2020-03-30
Updated lint.

PaulWoitaschek:
I'm not 100% sure if this is all correct, I followed the implementation at https://github.com/googlesamples/android-custom-lint-rules/tree/master/android-studio-3

### DIFF
--- a/conductor/build.gradle
+++ b/conductor/build.gradle
@@ -13,10 +13,6 @@ android {
     }
 }
 
-configurations {
-    lintChecks
-}
-
 dependencies {
     testImplementation rootProject.ext.junit
     testImplementation rootProject.ext.roboelectric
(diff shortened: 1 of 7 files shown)


conductor/build.gradle
```diff
@@ -17,29 +17,13 @@ android {
...
-    from(configurations.lintChecks) {
-        rename { 'lint.jar' }
-    }
-    into 'build/intermediates/lint/'
-}
-
-project.afterEvaluate {
-    def compileLintTask = project.tasks.find { it.name == 'compileLint' }
-    compileLintTask.dependsOn(copyLintJar)
+    lintChecks project(':conductor-lint')
```
DSteve595: You want `lintPublish` here (added in 3.4, which is why those 3.0 docs don't have it).

`lintChecks` = "use these lint checks to validate my project"

`lintPublish` = "use these lint checks to validate my project and projects dependent on my project"

DSteve595 (changes requested):
Other than the dependency config, LGTM

PR #591, opened 2020-04-04
Add platform transition module

### LINKED ISSUE #590: Keep platform Transitions for TransitionChangeHandler
I'd really like to hang on to the platform transitions. Two reasons for this:
1. Anecdotally, I've run into a few issues trying to use androidx transitions and them not playing nicely with overlays.
2. The new [material motion](https://material.io/develop/android/theming/motion/) transitions are based on the platform transitions, not the androidx versions.

This second point is probably the major one. (I can't currently find any example issues I mention in the first point).

My understanding is that currently, androidx transitions _don't_ bring anything new to the table in terms of features or bug fixes. I might be wrong, but they seem to _only_ backport transitions to pre API 21 devices.

With this in mind, could we reinstate `TransitionChangeHandler`? Or at least add them as a separate module alongside the androidx variant?

chris-horner:
Fixes #590. I thought about also porting to Kotlin, but I figure that's something for later, not now 😅


PaulWoitaschek:
I'm not sure about the naming. Maybe they should be called `conductor-transition-framework` and `conductor-transition-androidx`?

chris-horner:
Mmm, I do like those better. My concern was around renaming the current `androidx-transition` module to not break consumers, but I guess this version of Conductor hasn't officially released yet. I'll make the change.

chris-horner:
PR updated with new module and package names. Lemme know what you think. Happy to change to any suggestions you might have.

EricKuck:
Sorry for the tiny nit (really, I am), but I think `conductor-transition-platform` might be better. When I read “transition framework” it makes me think the module will be some kind of framework for making transitions. Does that make sense or am I crazy?

PaulWoitaschek:
Yes sounds good to me.

chris-horner:
PR updated with new names.
> Sorry for the tiny nit

Hitting shift+F6 is fine, don't worry about it!

EricKuck:
Thanks again!

PaulWoitaschek:
Imo we should revert this. 

https://mobile.twitter.com/ricknout/status/1257319370537070602

@erickuck cc

EricKuck:
I'm inclined to agree. @chris-horner after this news I think the reasons to continue supporting platform transitions are dropping to near-zero. Do you still see a need for it now?

chris-horner:
Nah, I reckon we can get rid of it now 🙂

Issue #592, opened 2020-04-11
Several controllers reattached


sdelaysam:
There is an issue when router is detached while executing animated push transactions.
In this case several routers in backstack have `needsAttach` flag set to true.
Obviously, next `rebindIfNeeded` will attach them all.

I'm gonna push PR with the test and fix.

PR #593, opened 2020-04-11
Bug 592: fix needsAttach flag on host detach

sdelaysam:
Fix for #592

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Router.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Router.java
@@ -632,12 +632,12 @@ public abstract class Router {
     public void prepareForHostDetach() {
         pendingControllerChanges.clear(); // rely on backstack based restoration in rebindIfNeeded
 
-        for (RouterTransaction transaction : backstack) {
-            if (ControllerChangeHandler.completeHandlerImmediately(transaction.controller().getInstanceId())) {
-                transaction.controller().setNeedsAttach(true);
+        if (!backstack.isEmpty()) {
+            RouterTransaction top = backstack.peek();
+            for (RouterTransaction transaction : backstack) {
+                ControllerChangeHandler.completeHandlerImmediately(transaction.controller().getInstanceId());
+                transaction.controller().prepareForHostDetach(transaction == top);
             }
-
-            transaction.controller().prepareForHostDetach();
         }
     }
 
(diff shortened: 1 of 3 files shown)


conductor/src/main/java/com/bluelinelabs/conductor/Router.java
```diff
@@ -632,12 +632,12 @@ public void onActivityDestroyed(@NonNull Activity activity) {
...
         pendingControllerChanges.clear(); // rely on backstack based restoration in rebindIfNeeded
 
-        for (RouterTransaction transaction : backstack) {
-            if (ControllerChangeHandler.completeHandlerImmediately(transaction.controller().getInstanceId())) {
-                transaction.controller().setNeedsAttach(true);
+        if (!backstack.isEmpty()) {
+            RouterTransaction top = backstack.peek();
+            for (RouterTransaction transaction : backstack) {
+                ControllerChangeHandler.completeHandlerImmediately(transaction.controller().getInstanceId());
+                transaction.controller().prepareForHostDetach(transaction == top);
```
EricKuck: It is possible for more than one controller to be attached at the same time, so this line will cause issues. Why couldn't we just rely on the logic that was already in `prepareForHostDetach` here (`needsAttach || attach`)?

sdelaysam: I wonder why backstack should have more than one attached controller? I was thinking that several controllers can be attached at the same time only if they belong to different routers.
Otherwise I would suggest to add the test for the expected behavior.

> Why couldn't we just rely on the logic that was already

I've faced the issue in my project:
When I push the controller with animated change handler, there is a moment when previous top controller is still attached and controller being pushed is marked as `needsAttach`.
If router is detached during the animation then both controllers are marked as `needsAttach` and both controllers get attached when router is rebound.

I covered this case with the test to demonstrate it and ensure its fixed.
https://github.com/bluelinelabs/Conductor/pull/593/files#diff-e4934720d381a89fbfe1e3da852c3440R496

EricKuck: There's a `removesFromViewOnPush` flag that would allow people to purposely leave more than one attached controller at any given time. One potential use of this is for dialog-style controllers where you still want to see the previous screen in the background.

Does calling `completeHandlerImmediately` like you did, but leaving the existing logic in `prepareForHostDetach` solve your issues, or is there still a gap there?

sdelaysam: No, unfortunately it doesn't work. (It was called in original version).
I also tried `performPendingControllerChanges()` instead of `pendingControllerChanges.clear()` - no luck as well.

> One potential use of this is for dialog-style controllers

Gotcha, need to consider this scenario in tests as well.

sdelaysam: I've pushed workaround for the use case mentioned above. Can you please review and reopen this conversation if it doesn't look solid enough.

PR #594, opened 2020-04-14
Adds ViewPager2 Adapter

EricKuck:
Note: This is currently lacking tests. Need to add these before merging.


MkhytarMkhoian:
Nice, thanks a lot

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+        currentPrimaryRouterPosition = position
+      }
+    }
+  }
+
+  @Parcelize
+  private data class SavedState(
+    val savedPagesKeys: List<Long>,
+    val savedPagesValues: List<Bundle>,
+    var savedPageHistory: List<Long>,
```
PaulWoitaschek: val

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+    var maxPagesToStateSave: Int
```
PaulWoitaschek: val

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+      savedPageHistory = savedPageHistory,
+      maxPagesToStateSave = maxPagesToStateSave
+    )
+  }
+
+  override fun restoreState(state: Parcelable) {
+    if (state !is SavedState) return
+
+    savedPages = LongSparseArray()
+    (0..state.savedPagesKeys.size).forEach { index ->
```
PaulWoitaschek: 0 unitl ..., this is one index too far.

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+   * the page that was state saved least recently will have its state removed from the save data.
+   */
+  open fun setMaxPagesToStateSave(maxPagesToStateSave: Int) {
+    require(maxPagesToStateSave >= 0) { "Only positive integers may be passed for maxPagesToStateSave." }
+    this.maxPagesToStateSave = maxPagesToStateSave
+    ensurePagesSaved()
+  }
+
+  private fun inferViewPager(recyclerView: RecyclerView): ViewPager2 {
+    val parent = recyclerView.parent
```
PaulWoitaschek:
``` 
   val parent = recyclerView.parent
    return parent as? ViewPager2? ?: error("Expected ViewPager2 instance. Got: $parent")
```

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterViewHolder.kt
```diff
@@ -0,0 +1,30 @@
...
+class RouterViewHolder private constructor(container: FrameLayout) : ViewHolder(container) {
+  val container: ChangeHandlerFrameLayout
+    get() = itemView as ChangeHandlerFrameLayout
+
+  var currentRouter: Router? = null
+  var currentItemPosition = 0
+  var currentItemId = 0L
+
+  companion object {
+    fun create(parent: ViewGroup): RouterViewHolder {
```
PaulWoitaschek: `operator fun invoke`

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterViewHolder.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.view.ViewGroup
+import android.widget.FrameLayout
+import androidx.core.view.ViewCompat
+import androidx.recyclerview.widget.RecyclerView.ViewHolder
+import com.bluelinelabs.conductor.ChangeHandlerFrameLayout
+import com.bluelinelabs.conductor.Router
+
+class RouterViewHolder private constructor(container: FrameLayout) : ViewHolder(container) {
+  val container: ChangeHandlerFrameLayout
```
PaulWoitaschek: Remove this and just put it in the construtor:
```kotlin
class RouterViewHolder
private constructor(
  val container: ChangeHandlerFrameLayout
) : ViewHolder(container) {
```

demo/src/main/java/com/bluelinelabs/conductor/demo/controllers/ViewPager2Controller.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    protected void onViewBound(@NonNull View view) {
+        super.onViewBound(view);
+        viewPager.setAdapter(pagerAdapter);
+        new TabLayoutMediator(tabLayout, viewPager, (tab, position) -> tab.setText("Page " + position)).attach();
+    }
+
+    @Override
+    protected void onDestroyView(@NonNull View view) {
+        if (!getActivity().isChangingConfigurations()) {
```
PaulWoitaschek, line 64: What is this good for?

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * An ViewPager2 adapter that uses Routers as pages
+ */
+abstract class RouterStateAdapter(private val host: Controller) :
+  RecyclerView.Adapter<RouterViewHolder>(), StatefulAdapter {
+
+  private var savedPages = LongSparseArray<Bundle>()
+  private var savedPageHistory = mutableListOf<Long>()
+  private var maxPagesToStateSave = Int.MAX_VALUE
+  private val visibleRouters = SparseArray<Router>()
```
PaulWoitaschek: val

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+    return position.toLong()
+  }
+
+  override fun setHasStableIds(hasStableIds: Boolean) {
+    throw UnsupportedOperationException("Stable Ids are required for the adapter to function properly")
+  }
+
+  override fun saveState(): Parcelable {
+    return SavedState(
+      savedPagesKeys = (0..savedPages.size()).map { savedPages.keyAt(it) },
```
PaulWoitaschek: 0 until size

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,190 @@
...
+      savedPagesValues = (0..savedPages.size()).map { savedPages.valueAt(it) },
```
PaulWoitaschek: 0 until size

PaulWoitaschek (changes requested):
I left some reviews.

This needs tests and it's crashing on configuration change due to various issues.

Also I'd hide all the implementation stuff (adapter & viewHolder) and just expose a top level function:

```kotlin
typealias ConfigureRouter = (router: Router, position: Int) -> Unit

fun routerStateAdapter(
  host: Controller,
  maxPagesToStateSave: Int = Int.MAX_VALUE,
  configureRouter: ConfigureRouter
): RecyclerView.Adapter<RecyclerView.ViewHolder> {
  TODO()
}
```

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
DSteve595: Alternatively, `0..state.savedPagesKeys.lastIndex`

PaulWoitaschek: indices.forEach :)

PaulWoitaschek: I tend to not use these android collections at all these days because you lose the kotlin collection stuff and it doesn't matter here anyways (this think holding multiple views, here a few ints are just nuts)

EricKuck: oof…

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,221 @@
...
+      savedPages.put(holder.currentItemId, savedState)
+
+      savedPageHistory.remove(holder.currentItemId)
+      savedPageHistory.add(holder.currentItemId)
+
+      ensurePagesSaved()
+
+      visibleRouters.remove(holder.currentItemPosition)
+    }
+
```
alanvan-agl: @EricKuck if `detachRouter` is called after `attachRouter`, for example in the case of underlying data changes, then `visibleRouters` will be inaccurately updated. Suggestion:
`            if (visibleRouters[holder.currentItemPosition] == router) {
                visibleRouters.remove(holder.currentItemPosition)
            }`

conductor-modules/viewpager2/src/main/java/com/bluelinelabs/conductor/viewpager/RouterStateAdapter.kt
```diff
@@ -0,0 +1,221 @@
...
+  private fun detachRouter(holder: RouterViewHolder) {
+    holder.currentRouter?.let { router ->
+      router.prepareForHostDetach()
+
+      val savedState = Bundle()
+      router.saveInstanceState(savedState)
+      savedPages.put(holder.currentItemId, savedState)
+
+      savedPageHistory.remove(holder.currentItemId)
+      savedPageHistory.add(holder.currentItemId)
```
alanvan-agl: @EricKuck remove then add? should it just be `savedPageHistory.add(holder.currentItemId)`

sdelaysam:
I'm playing with this adapter and noticed that views are not actually removed from container when swiped out.

PaulWoitaschek:
Nice @EricKuck :tada:

sdelaysam:
@EricKuck @PaulWoitaschek  there is still an issue with current implementation
Check my comment above.
To verify, just change the code of `ChildController.onViewBound` to
```
@Override
protected void onViewBound(@NonNull View view) {
    super.onViewBound(view);

    tvTitle.setText(String.valueOf(System.currentTimeMillis()));
}
```

and swipe back and forth.

EricKuck:
Thanks for the reminder @sdelaysam. Sorry I missed this the first time through. I just pushed a fix for this issue on develop.

sdelaysam:
@EricKuck there is another issue with adapter. Child controllers don't restore their states.
This what happens when I push controller over `ViewPager2Controller` and press back.
```
// push
Child #0 (Swipe to see more) saves instance state
Child #0 (Swipe to see more) view destroyed
Child #0 (Swipe to see more) destroyed

// press back
Top view destroyed
Top destroyed
```

Same scenario on `ViewPagerController`
```
// push
Child #0 (Swipe to see more) saves instance state
Child #0 (Swipe to see more) view destroyed
Child #0 (Swipe to see more) destroyed

// press back
Top view destroyed
Top destroyed
Child #0 (Swipe to see more) restores instance state // <-- 
```

I was also surprised that child controllers are fully destroyed on push.
In fact their parent (ViewPagerController/ViewPager2Controller) is in the backstack.

EricKuck:
@sdelaysam I see that the logs for this are different, but I think that boils down to the difference between VP1 and VP2. Because VP2 doesn't have a consistent "this item is going to be destroyed" callback that VP1 had, it ends up over-saving instance states to be safe, but it doesn't fail to restore them. Swipe enough that a controller is fully destroyed, then swipe back to it - it should get its instance state restored just fine.

sdelaysam:
> it should get its instance state restored just fine 

it should but it does not, you can easily check that in the demo app.
I guess there is a bug in state restoring logic.

EricKuck:
Ah sorry I totally misread your repro steps. I see the problem now. Thanks for the heads up!

Wrywulf:
@EricKuck 
I also noticed that "abandoned" controller views are not detached from the `RouterViewHolder`s layout container when swiping back -> forward -> back -> forward etc etc.
The number of `RouterViewHolder.container` children just keeps increasing. I also verified this using LayoutInspector.

```
 private fun attachRouter(holder: RouterViewHolder, position: Int) {
        val itemId = getItemId(position)

        val router = host.getChildRouter(holder.container, "$itemId")  
...
```
If the combination of `holder.container` and `"$itemId"` results in a new Router being created, it just adds to the holder's container.

If the holder is being re-used, this just adds up over time. Only when the holder is recycled does it clean up (removing the *current* router).

Perhaps it doesn't immediately cause any negative effects on normal usage, but I noticed that when talk-back is enabled it becomes totally confused about what views to focus etc.

My current workaround is to simply remove any existing routers from a "dirty" `RouterViewHolder` before proceeding:

```
 private fun attachRouter(holder: RouterViewHolder, position: Int) {
        val itemId = getItemId(position)
        // dirty holder - clean up!
        holder.currentRouter?.let {  host.removeChildRouter(it) }

        val router = host.getChildRouter(holder.container, "$itemId")
...
```

PR #602, opened 2020-05-11
Migrated to AndroidX Fragments

KirillMalafey:
Motivation:
1) android.app.Fragment is deprecated;
2) Consistent behavior across different api versions;
3) I use Conductor in my project and noticed probably unexpected behavior when working with runtime permissions. With androidx.Fragment or AppCompatActivity if you try to start another one request with the same id, previous request will be automatically cancelled by Android. But android.app.Fragment does not do this (at least on Android 7 and 8), it invokes onRequestPermissionsResult() method twice.


KirillMalafey:
@EricKuck @PaulWoitaschek

ursusursus:
We need this! I was burned by this exact issue.

PR #644, opened 2021-06-18
Add basic support for compose.

PaulWoitaschek:


conductor/src/main/java/com/bluelinelabs/conductor/internal/OwnViewTreeLifecycleAndRegistry.kt
```diff
@@ -0,0 +1,118 @@
...
+        })
+    }
+
+    override fun getLifecycle(): Lifecycle = lifecycleRegistry
+
+    override fun getSavedStateRegistry(): SavedStateRegistry =
+        savedStateRegistryController.savedStateRegistry
+
+    companion object {
+        fun own(target: Controller) {
```
DSteve595: Would be nice to have a comment explaining what/why this is

DSteve595 (approved):
Nice!

Is there a long-term vision for Compose integration? A controller class with `@Composable fun content()` is all that's really _needed_. But in the future, if Compose adds some facilities for shared element transitions, I imagine it might be desirable to remove the View backing in favor of pure Composable content. Does that make sense to do? If so, would it be a "fork" of Conductor, or another more of operation, or something else?

conductor/src/main/java/com/bluelinelabs/conductor/internal/OwnViewTreeLifecycleAndRegistry.kt
PaulWoitaschek: What do you mean? What the class is for? It's not pubic api (at least not on kotlin) 😉

DSteve595: Yeah, what the class is for. It may not be public, but I have _somewhat_ of an understanding of Conductor's internals, and it isn't clear to me what it's achieving. Feel like it might be hard to remember context a year from now.

EricKuck (approved):

conductor/src/main/java/com/bluelinelabs/conductor/internal/OwnViewTreeLifecycleAndRegistry.kt
chris-hatton: I'm a Conductor user who just tried naively using a `ComposeView` in my `Controller` and hit lifecycle related exceptions.  Searching led me here; but neither this PR, or anything else I can find, leaves me with a clue how to make Compose place nicely with Conductor...

> This class sets the [ViewTreeLifecycleOwner] and [ViewTreeSavedStateRegistryOwner] which is necessary for Jetpack Compose. By setting these...

How?

> ...the view state restoration and compose lifecycle play together with the lifecycle of the [Controller].

Yes, that's definitely what I want 😄  
Example, please?

chris-hatton: > This class sets the [ViewTreeLifecycleOwner] and [ViewTreeSavedStateRegistryOwner] which is necessary for Jetpack Compose. By setting these
How!??
> the view state restoration and compose lifecycle play together with the lifecycle of the [Controller].
Example, please.

PaulWoitaschek: Can you open a bug report with a complete stack trace and the conductor & compose versions you are using?

PR #678, opened 2022-07-28
Fix NPE when removing view reference

MarioNoll:
After upgrading from `3.1.5` to `3.1.6`, we can see users running into a NPE, introduced [here](https://github.com/bluelinelabs/Conductor/compare/3.1.5...3.1.6#diff-45f17ca3409e4f5b7bde2067f79f99c7fff0345979fbb8d2aece39a7c2cdddb2R1085).

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
@@ -1082,7 +1082,7 @@ public abstract class Controller {
     final View inflate(@NonNull ViewGroup parent) {
         if (view != null && view.getParent() != null && view.getParent() != parent) {
             detach(view, true, false);
-            removeViewReference(view.getContext());
+            removeViewReference(view != null ? view.getContext() : null);
         }
 
         if (view == null) {


conductor/src/main/java/com/bluelinelabs/conductor/Controller.java
```diff
@@ -1082,7 +1082,7 @@ private void removeViewReference(@Nullable Context context) {
     final View inflate(@NonNull ViewGroup parent) {
         if (view != null && view.getParent() != null && view.getParent() != parent) {
             detach(view, true, false);
-            removeViewReference(view.getContext());
+            removeViewReference(view != null ? view.getContext() : null);
```
PaulWoitaschek, line 1085: Wouldn't a more correct behavior be, to capture the view locally in the beginning of the method body?

EricKuck: This is a good call. Doesn't matter in this case because the context is only used when the controller is being destroyed and it will never be getting destroyed during `inflate`. But for the sake of correctness and avoiding future footguns I'll update it to work this way.

PR #680, opened 2022-11-29
Update ancestor change listeners to prevent memory leak

EricKuck:
OwnViewTreeLifecycleAndRegistry adds listeners for each of its owning controller's ancestors. The links between controllers may be broken by the time the callbacks happen, meaning it will be unable to remove those listeners. This results in a memory leak. In order to prevent associations like this, this PR adds the concept of a global static listener that all controllers can subscribe to in order to be made aware of changes in the controllers they're interested in.


brycecrookston (approved):

fpsulli3 (approved):

brycecrookston (changes requested):
tests broken

brycecrookston (approved):

Issue #389, opened 2017-11-30
ConcurrentModificationException iterating Backstack


toddhodes:
Hi!  I'm seeing the following.  The flow is from a conductor activity, to a dialog, the calling out to the SMS app, and then when the user hits the back button from the native SMS app, we get the following:

```
java.util.ConcurrentModificationException                                                            
   at java.util.ArrayDeque$DeqIterator.next(ArrayDeque.java:608)                                     
   at com.bluelinelabs.conductor.Router.onActivityStarted(Router.java:532)                           
   at com.bluelinelabs.conductor.internal.LifecycleHandler.onActivityStarted(LifecycleHandler.java:334)
   at android.app.Application.dispatchActivityStarted(Application.java:205)                          
   at android.app.Activity.onStart(Activity.java:1150)                                               
   at android.support.v4.app.FragmentActivity.onStart(FragmentActivity.java:564)                     
   at android.support.v7.app.AppCompatActivity.onStart(AppCompatActivity.java:177)                   
   at com.locationlabs.cni.mvp.BaseControlsActivity.onStart(BaseControlsActivity.java:51)            
   at com.locationlabs.cni.contentfiltering.AppControlsActivity.onStart(AppControlsActivity.java:117)
   at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1237)                     
   at android.app.Activity.performStart(Activity.java:6268)                                          
   at android.app.Activity.performRestart(Activity.java:6314)                                        
   at android.app.Activity.performResume(Activity.java:6319)                                         
   at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3092)                     
   at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3134)                      
   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1388)                           
   at android.os.Handler.dispatchMessage(Handler.java:102)                                           
   at android.os.Looper.loop(Looper.java:148)                                                        
   at android.app.ActivityThread.main(ActivityThread.java:5417)                                      
   at java.lang.reflect.Method.invoke(Native Method)                                                 
   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)                
   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

can you make the Router's Backstack iterator thread-safe, or, provide any other advice?  as a workaround, I'm simply catching it and moving on...

```
   @Override
[...]

PR #685, opened 2023-02-17
Fix lifecycle crash - Change backstack from ArrayDeque to LinkedBlockingDeque 

Bryan-Lamb:
**Affected Version:** 
Conductor 3.2.0

**Context:**
In a typical Controller, we begin observing events from our ViewModels (using channels/flows) in the `LifecycleListener.postAttach()` callback which often result in a navigation call such as `pushController()` or `setRoot()`.

In particular with `setRoot()` we've been encountering an interesting crash for quite some time and after investigation narrowed down a reproducible scenario to be:
1. Slow API call triggered, where a successful response results in `setRoot()` being called
2. While API call is inflight, the App is backgrounded
3. App is resumed, postAttach is triggered and calls `setRoot()` while the Backstack is still being iterated over by the `LifecyleHandler`

There appears to be an [existing issue](https://github.com/bluelinelabs/Conductor/issues/389) with similar circumstances

```
Exception java.util.ConcurrentModificationException:
  at java.util.ArrayDeque$DeqIterator.next (ArrayDeque.java:631)
  at com.bluelinelabs.conductor.RouterTransaction.controller (RouterTransaction.java)
  at com.bluelinelabs.conductor.internal.LifecycleHandler.onActivityStarted (LifecycleHandler.java)
  at android.app.Application.dispatchActivityStarted (Application.java:405)
  at android.app.Activity.dispatchActivityStarted (Activity.java:1406)
  at android.app.Activity.onStart (Activity.java:1922)
  at androidx.fragment.app.FragmentController.dispatchActivityCreated (FragmentController.java)
  at androidx.appcompat.app.AppCompatActivity.onStart (AppCompatActivity.java)
  at android.app.Instrumentation.callActivityOnStart (Instrumentation.java:1455)
  at android.app.Activity.performStart (Activity.java:8315)
  at android.app.ActivityThread.handleStartActivity (ActivityThread.java:4136)
  at android.app.servertransaction.TransactionExecutor.performLifecycleSequence (TransactionExecutor.java:221)
  at android.app.servertransaction.TransactionExecutor.cycleToPath (TransactionExecutor.java:201)
  at android.app.servertransaction.TransactionExecutor.executeLifecycleState (TransactionExecutor.java:173)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:97)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2443)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8751)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:571)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1135)
```

**Suggested Change**
Make Backstack synchronised by changing it's type from an `ArrayDeque` to a `LinkedBlockingDeque`

I've added a test to simulate this scenario which will fail with the previous `ArrayDeque` implementation, confirmed all tests are passing, but interested to know if there may be other unintended consequences of this change.

### DIFF
--- a/conductor/src/main/java/com/bluelinelabs/conductor/Backstack.kt
+++ b/conductor/src/main/java/com/bluelinelabs/conductor/Backstack.kt
@@ -2,10 +2,11 @@ package com.bluelinelabs.conductor
 
 import android.os.Bundle
 import java.util.*
+import java.util.concurrent.LinkedBlockingDeque
 
 internal class Backstack : Iterable<RouterTransaction> {
 
-  private val backstack: Deque<RouterTransaction> = ArrayDeque()
+  private val backstack: Deque<RouterTransaction> = LinkedBlockingDeque()
 
   val isEmpty: Boolean get() = backstack.isEmpty()
 
--- a/conductor/src/test/java/com/bluelinelabs/conductor/ReattachCaseTests.kt
+++ b/conductor/src/test/java/com/bluelinelabs/conductor/ReattachCaseTests.kt
@@ -2,7 +2,9 @@ package com.bluelinelabs.conductor
 
 import android.os.Bundle
 import android.os.Looper
+import android.view.View
 import com.bluelinelabs.conductor.Conductor.attachRouter
+import com.bluelinelabs.conductor.Controller.LifecycleListener
 import com.bluelinelabs.conductor.internal.LifecycleHandler
 import com.bluelinelabs.conductor.util.ActivityProxy
 import com.bluelinelabs.conductor.util.AttachFakingFrameLayout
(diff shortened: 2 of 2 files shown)


RedaMohsen:
The same crash randomly happened to my app on Android 12 as reported by crashlytics. Hopefully this is a valid solution and the newest version will not have this issue.

Bryan-Lamb:
Interested to hear any thoughts you may have on this @EricKuck

daewin:
+1 We've observed this crash happening for a while now too

EricKuck:
Thanks for taking a look at this one. I'm not sure this is the fix we need though. The test actually passes without any changes to the backstack. Additionally, is blocking the right way to handle this? If all changes are happening on the main thread, that could result in a deadlock.

PR #693, opened 2024-03-30
Refine highlighting and error reporting in Lint rules

jamesonwilliams:
# What is this?

Previously, the Lint rules would underline the entire class in Android Studio. This was visible, but somewhat distracting, and could cover up other Lint issues.

This commit makes two changes:
  1. Only highlight the offending class/constructor in the Android Studio
  2. When the Lint violation relates to a constructor as opposed to a class, highlight the constructor itself.

Thanks to @scana for identifying the problem and solution space 

# Testing

To simplify testing, I created a copy of the rule called `ValidConductorController`, so I could A/B them using `@SuppressLint("...")`. 

## Before
This is how the Conductor Lint rules currently materialize, over the entire "`node`":
Screenshot 2024-03-29 at 8 52 22 PM

## After

With the updates, the squiggly error line only shows up under the constructor or class name, as appropriate:


jamesonwilliams:
cc: @scana

EricKuck (approved):

